PR #7, opened 2025-05-16
Add tool mem0_memory

deshraj:
## Description
Adds support for `mem0_memory` tool to use memory with agents so that they can remember past states and personalize the experience of the users.

## Type of Change
- [x] New Tool

## Testing
Yes, tested locally and added tests as well.

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

### DIFF
--- a/tests/test_mem0.py
+++ b/tests/test_mem0.py
@@ -1,0 +1,411 @@
+"""
+Tests for the memory tool using the Agent interface.
+"""
+
+import os
+from unittest.mock import MagicMock, patch
+
+import pytest
+from strands import Agent
+from strands.sdk.types.tools import ToolUse
+from strands_tools import mem0_memory
+from strands_tools.mem0_memory import Mem0ServiceClient
+
+
+@pytest.fixture
+def agent():
+    """Create an agent with the memory tool loaded."""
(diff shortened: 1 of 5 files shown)


cagataycali (commented):
Looks like one more small detail left:

```
tests/test_mem0.py:10: in <module>
    from strands.sdk.types.tools import ToolUse
E   ModuleNotFoundError: No module named 'strands.sdk'
```

tests/test_mem0.py
```diff
@@ -0,0 +1,411 @@
+"""
+Tests for the memory tool using the Agent interface.
+"""
+
+import os
+from unittest.mock import MagicMock, patch
+
+import pytest
+from strands import Agent
+from strands.sdk.types.tools import ToolUse
```
deshraj:
```suggestion
from strands.types.tools import ToolUse
```

cagataycali (approved):
lgtm!

tests/test_mem0.py
cagataycali: Looks like still this change is not in the PR 🤔

deshraj: There is some issue with Github. It shows old commit.

PR #16, opened 2025-05-19
Add different configurations to Mem0 tool

deshraj:
## Description
Adding support for other configurations in mem0 tool so that it is easier for developers to integrate this tool.

The Mem0 Memory Tool supports three different backend configurations:

1. **Mem0 Platform** (Recommended for production):
   - Uses the Mem0 Platform API for memory management
   - Requires a Mem0 API key

2. **OpenSearch** (Recommended for AWS environments):
   - Uses OpenSearch as the vector store backend
   - Requires AWS credentials and OpenSearch configuration

3. **FAISS** (Default for local development):
   - Uses FAISS as the local vector store backend
   - Requires faiss-cpu package for local vector storage


README.md
```diff
@@ -238,9 +238,31 @@ These variables affect multiple tools:
 
 #### Mem0 Memory Tool
 
-| Environment Variable | Description | Default |
-|----------------------|-------------|---------|
-| OPENSEARCH_HOST | OpenSearch Host URL | None |
+The Mem0 Memory Tool supports three different backend configurations:
+
+1. **Mem0 Platform** (Recommended for production):
```
Unshure: nit: Can you remove this `(Recommended for production)` comment

src/strands_tools/mem0_memory.py
```diff
@@ -402,12 +487,24 @@ def format_history_response(history: List[Dict]) -> Panel:
...
+    if isinstance(results, list):
+        if not results:
+            return Panel("No memories stored.", title="[bold yellow]No Memories Stored", border_style="yellow")
+        memories = results
+    # Handle OpenSearch/FAISS response format (dict with results key)
+    else:
+        if not results.get("results"):
+            return Panel("No memories stored.", title="[bold yellow]No Memories Stored", border_style="yellow")
+        memories = results.get("results", [])
+
```
Unshure: It looks like we have many checks throughout the code to see if the results are a dictionary or a list. Can we consolidate this logic, so right after getting `results`, we transform it to a list, and then process it? Seems like that would simplify the logic everywhere.

src/strands_tools/mem0_memory.py
```diff
@@ -291,8 +356,16 @@ def format_get_response(memory: Dict) -> Panel:
 
 def format_list_response(memories: Dict) -> Panel:
```
Unshure: Can you update the type of this input parameter to match what is expected?

src/strands_tools/mem0_memory.py
```diff
@@ -596,6 +709,7 @@ def mem0_memory(tool: ToolUse, **kwargs: Any) -> ToolResult:
             raise ValueError(f"Invalid action: {action}")
 
     except Exception as e:
+        traceback.print_exc()
```
Unshure: Can you remove this?

src/strands_tools/mem0_memory.py
```diff
@@ -185,45 +188,107 @@ def __init__(self, config: Optional[Dict] = None):
...
+    def _initialize_client(self, config: Optional[Dict] = None) -> Any:
+        """Initialize the appropriate Mem0 client based on environment variables.
+
+        Args:
+            config: Optional configuration dictionary to override defaults.
+
+        Returns:
+            An initialized Mem0 client (MemoryClient or Mem0Memory instance).
+        """
+        if os.environ.get("MEM0_API_KEY"):
```
Unshure, line 206: Can you add debug logging around which client was chosen?

Unshure:
Thank you for your contributions here, we are interested in expanding the capabilities of the `mem0_memory` tool! Ive left a few comments above, and I will work on testing these changes myself as well.

deshraj:
Thanks for the comments @Unshure. Incorporated all your suggestions.

Unshure:
Hi @deshraj, thanks for following up with those fixes!

I got the chance to pull your change locally and test it out and ran into a couple of errors during my testing:
- `action: list` failed with the error: `'list' object has no attribute 'id'`. This seems like somewhat transient behavior, but mostly in the faiss implementation. Did you run into this?
- Looks like the `get_history` method doesn't actually exist, and its really supposed to be `history`. This was introduced outside of your pr, but I was curious if this was an issue you ran into as well?
    - documentation link: https://docs.mem0.ai/open-source/python-quickstart#memory-history

Only blocker right now is the `list` action transiently failing. If you see this as well, we will have to come up with a work around for it.

deshraj:
Hey @Unshure, thanks for the review. I tried again and it seems to work fine with all the configurations. Can you please try the [demo here](https://gist.github.com/deshraj/3b63ecedbef5561ad07cd49865ccf198) and see if you run into the same issue? It would help if you can share the full traceback.

Also, I have fixed the issue related to history. Thanks for pointing it out.

Unshure:
Thanks for fixing the `history` issue, appreciate that!

As for the `action: list` issue I mentioned, here is the output that I see when running [your linked example](https://gist.github.com/deshraj/3b63ecedbef5561ad07cd49865ccf198):
```
> Show off your memory tool
I'd be happy to demonstrate how the memory tool works! This feature allows me to remember important information about you and recall it later in our conversations.

Let me show you a few examples of what the memory tool can do:

1. First, let me check if there are any existing memories stored for you:
Tool #1: mem0_memory
/home/ncclegg/.local/share/mise/installs/python/3.10.17/lib/python3.10/site-packages/faiss/loader.py:49: DeprecationWarning: numpy.core._multiarray_umath is deprecated and has been renamed to numpy._core._multiarray_umath. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core._multiarray_umath.__cpu_features__.
  from numpy.core._multiarray_umath import __cpu_features__
INFO:faiss.loader:Loading faiss with AVX2 support.
INFO:faiss.loader:Successfully loaded faiss with AVX2 support.
INFO:botocore.credentials:Found credentials in shared credentials file: ~/.aws/credentials
INFO:mem0.vector_stores.faiss:Loaded FAISS index from /tmp/mem0_384_faiss/mem0.faiss with 1 vectors
INFO:mem0.vector_stores.faiss:Loaded FAISS index from /home/ncclegg/.mem0/migrations_faiss/mem0migrations.faiss with 1 vectors
Traceback (most recent call last):
  File "/workplace/ncclegg/tools/src/strands_tools/mem0_memory.py", line 619, in mem0_memory
    memories = client.list_memories(tool_input.get("user_id"), tool_input.get("agent_id"))
  File "/workplace/ncclegg/tools/src/strands_tools/mem0_memory.py", line 319, in list_memories
    return self.mem0.get_all(user_id=user_id, agent_id=agent_id)
  File "/home/ncclegg/.local/share/mise/installs/python/3.10.17/lib/python3.10/site-packages/mem0/memory/main.py", line 546, in get_all
    all_memories_result = future_memories.result()
  File "/home/ncclegg/.local/share/mise/installs/python/3.10.17/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/home/ncclegg/.local/share/mise/installs/python/3.10.17/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/home/ncclegg/.local/share/mise/installs/python/3.10.17/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/ncclegg/.local/share/mise/installs/python/3.10.17/lib/python3.10/site-packages/mem0/memory/main.py", line 582, in _get_all_from_vector_store
    id=mem.id,
AttributeError: 'list' object has no attribute 'id'
╭───────────────────────────────────────────── ❌ Memory Operation Error ─────────────────────────────────────────────╮
│ 'list' object has no attribute 'id'                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
It looks like there might be an issue with retrieving the existing memories. Let me store a new memory to demonstrate how this works:
```

I get this issue when using faiss-cpu and with opensearch, but I dont get it when using the mem0 integration. Please let me know what you think.

pyproject.toml
```diff
@@ -62,7 +62,7 @@ dev = [
     "pytest>=8.0.0,<9.0.0",
     "ruff>=0.4.4,<0.5.0",
     "responses>=0.6.1,<1.0.0",
-    "mem0ai>=0.1.99,<1.0.0",
+    "mem0ai>=0.1.102,<1.0.0",
```
deshraj:
```suggestion
    "mem0ai>=0.1.104,<1.0.0",
```

deshraj:
Thanks @Unshure for the review. Seems like the issue was with our `mem0ai` python package which is fixed now. I have updated the version of the package here.

Can you please test again?

Unshure:
The package update worked for me! Thanks for fixing this, Approved!

PR #53, opened 2025-05-26
fix(file_read): remove redundant base64 encoding of file content

### LINKED ISSUE #26: [BUG] The error won't go away without starting the agent. Error: An error occurred (ValidationException) when calling the ConverseStream operation: The detected filetype is PLAIN_TEXT, but the provided filetype was PDF.
### Checks

### Strands Version

0.5.2

### Python Version

3.12.10

### Operating System

macOS 15.4.1

### Installation Method

pip

### Steps to Reproduce

Error: An error occurred (ValidationException) when calling the ConverseStream operation: The detected filetype is PLAIN_TEXT, but the provided filetype was PDF.

This won't go away until I restart the agent. Every consecutive prompt is resulting in this error as I tried using file_read tool with pdf. 

### Expected Behavior

After the error normal prompt processing should proceed.

### Actual Behavior

The error above

### Additional Context

_No response_

### Possible Solution

_No response_

### Related Issues

_No response_

awalcutt:
## Description
The AWS SDK automatically base64 encodes binary content when sending requests, but the file_read tool was also encoding it, resulting in double-encoding. According to the Bedrock API documentation: "If you use an AWS SDK, you don't need to encode the bytes in base64."

This change fixes the issue by removing the extra base64 encoding step and letting the bedrock client handle it.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse_stream.html

## Related Issues
Fixes #26 

## Documentation PR
N/A

## Type of Change
- [X] Bug fix
- [ ] New Tool
- [ ] Breaking change

## Testing

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`
* installing in my private application

## Checklist

### DIFF
--- a/src/strands_tools/file_read.py
+++ b/src/strands_tools/file_read.py
@@ -96,7 +96,6 @@ agent.tool.file_read(
 See the file_read function docstring for more details on modes and parameters.
 """
 
-import base64
 import glob
 import json
 import os
@@ -188,13 +187,12 @@ def create_document_block(
             name_uuid = str(uuid.uuid4())[:8]
             neutral_name = f"{os.path.splitext(base_name)[0]}-{name_uuid}"
 
-        # Read and encode file content
+        # Read file content
         with open(file_path, "rb") as f:
             content = f.read()
-            encoded = base64.b64encode(content).decode("utf-8")
 
         # Create document block
-        return {"name": neutral_name, "format": format, "source": {"bytes": encoded}}
+        return {"name": neutral_name, "format": format, "source": {"bytes": content}}
 
     except Exception as e:
         raise Exception(f"Error creating document block for {file_path}: {str(e)}") from e


arronbailiss:
Hi @awalcutt, thank you for opening this PR.

How does this work with other model providers? Strands supports various model providers like Bedrock, Anthropic, LiteLLM, Llama API, Ollama, OpenAI, and custom providers.

awalcutt:
@awsarron Good question. I didn't look too deep into that because the docstring for this functions says ["Create a Bedrock document block from a file."](https://github.com/strands-agents/tools/blob/main/src/strands_tools/file_read.py#L164) 

[Anthropic Provider](https://github.com/strands-agents/sdk-python/blob/main/src/strands/models/anthropic.py#L102) is also separately encoding the content so this should fix that case as well.

As far as I can tell the other providers do not have specific handling for `document` content blocks but fall back on `json.dumps(content)` which will throw an error if bytes are passed in.

The [DocumentContent type](https://strandsagents.com/0.1.x/api-reference/types/#strands.types.media.DocumentContent) explicitly calls for `bytes` rather than `str` (via `DocumentSource`).

So... should `DocumentContent` contain bytes or an encoded string?

If bytes model providers must explicitly handle `document` blocks by throwing an error, filtering them out, or explicitly encoding them. Documentation should clarify that this is required for any model provider implementation. SDK can provide helper functions. I think this option is the most flexible to emerging/custom model provider implementations and is a cleaner interface between the responsibility of tools and model providers.

If encoded string model providers that need bytes (e.g. `Bedrock`) are responsible for decoding document blocks or configuring the underlying client to prevent additional encoding. Model providers that want a different encoding (encryption? compression?) may need to decode. Personally I find this interface a bit awkward but I recognize that practically speaking most providers are going to expect base64 encoding. There is a potential this is wasteful (of tokens and therefore money) if a provider does not support documents and gets sent a bunch of base64 encoded content that it can't use.

Thoughts?

pgrayy:
Hello @awalcutt, thank you for putting together this PR. After discussing with the team, we decided on the following:

1. Add document content handling in the OpenAI and LiteLLM model providers (https://github.com/strands-agents/sdk-python/pull/138)
   - As with Bedrock and Anthropic, we are going to handle the base64 encoding on behalf of customers and so the change you make here will still be valid.
 2. Throw an exception in the LlamaAPI and Ollama model providers when document blocks are passed in (PR pending).
    - These providers do not support document blocks and so having an explicit exception will be better than letting `json.dumps` fail.

Once these 2 changes are in, I would be okay merging your fix. Thank you for your patience.

PR #82, opened 2025-06-12
fix: fix non-interactive mode support to shell tool & terminate logic

JackYPCOnline:
## Description
fix the bug that non-interactive mode run infinitely in integration test

- Dual-Mode Operation: Added comprehensive support for both interactive and non-interactive execution through a new non_interactive_mode parameter
- Process Group Management: Improved command execution with better process isolation using os.setsid() for cleaner termination
- Terminal State Safety: Enhanced terminal settings handling to prevent issues in headless environments
- Signal Handling: Improved process termination with proper process group signaling
- Simplified API: Streamlined command normalization for more consistent behavior
- Update unit tests

## Related Issues

## Documentation PR

## Type of Change
- [ x] Bug fix
- [ ] New Tool
- [ ] Breaking change

## Testing
Yes. Unit tests and start strands agent.

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

## Checklist
- [ x] I have added tests that prove my fix is effective or my feature works

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ cd tools
 
 # Create and activate virtual environment
 python3 -m venv .venv
-source venv/bin/activate  # On Windows: venv\Scripts\activate
+source .venv/bin/activate  # On Windows: venv\Scripts\activate
 
 # Install in development mode
 pip install -e ".[dev]"
@@ -95,9 +95,9 @@ Below is a comprehensive table of all available tools, how to use them with an a
 | file_read | `agent.tool.file_read(path="path/to/file.txt")` | Reading configuration files, parsing code files, loading datasets |
 | file_write | `agent.tool.file_write(path="path/to/file.txt", content="file content")` | Writing results to files, creating new files, saving output data |
 | editor | `agent.tool.editor(command="view", path="path/to/file.py")` | Advanced file operations like syntax highlighting, pattern replacement, and multi-file edits |
-| shell | `agent.tool.shell(command="ls -la")` | Executing shell commands, interacting with the operating system, running scripts |
+| shell* | `agent.tool.shell(command="ls -la")` | Executing shell commands, interacting with the operating system, running scripts |
 | http_request | `agent.tool.http_request(method="GET", url="https://api.example.com/data")` | Making API calls, fetching web data, sending data to external services |
-| python_repl | `agent.tool.python_repl(code="import pandas as pd\ndf = pd.read_csv('data.csv')\nprint(df.head())")` | Running Python code snippets, data analysis, executing complex logic with user confirmation for security |
+| python_repl* | `agent.tool.python_repl(code="import pandas as pd\ndf = pd.read_csv('data.csv')\nprint(df.head())")` | Running Python code snippets, data analysis, executing complex logic with user confirmation for security |
 | calculator | `agent.tool.calculator(expression="2 * sin(pi/4) + log(e**2)")` | Performing mathematical operations, symbolic math, equation solving |
 | use_aws | `agent.tool.use_aws(service_name="s3", operation_name="list_buckets", parameters={}, region="us-west-2")` | Interacting with AWS services, cloud resource management |
 | retrieve | `agent.tool.retrieve(text="What is STRANDS?")` | Retrieving information from Amazon Bedrock Knowledge Bases |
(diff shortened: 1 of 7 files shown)


JackYPCOnline:
also tested with integration test with different interactive mode and termination.

src/strands_tools/shell.py
```diff
@@ -235,41 +240,43 @@ def execute_with_pty(self, command: str, cwd: str) -> Tuple[int, str, str]:
...
-                except OSError as e:
-                    logger.debug(f"Error waiting for process: {e}")
-                    exit_code = 1
-
+                    if os.WIFEXITED(status):
+                        exit_code = os.WEXITSTATUS(status)
+                    else:
+                        exit_code = -1  # Process was terminated by a signal
+                except OSError:
+                    exit_code = -1  # waitpid failed
```
zastrowm, line 264: Before we had an exit_code of 1 - was this changed because it's more correct?

tests/test_shell.py
```diff
@@ -947,21 +933,21 @@ def test_execute_single_command():
...
         cmd_dict = {"command": "git status", "timeout": 30}
-        result = shell.execute_single_command(cmd_dict, "/tmp", 10)
+        result = shell.execute_single_command(cmd_dict, "/tmp", 10, non_interactive_mode=False)
         assert result["command"] == "git status"
         assert result["options"] == cmd_dict
 
         # Test with exception
         mock_execute.side_effect = Exception("Test error")
-        result = shell.execute_single_command("failing command", "/tmp", 10)
+        result = shell.execute_single_command("failing command", "/tmp", 10, non_interactive_mode=False)
```
zastrowm, line 959: Do we have unit tests for the `non_interactive_mode=True` case?

src/strands_tools/shell.py
```diff
@@ -184,17 +184,18 @@ def __init__(self, timeout: int = None) -> None:
...
-
+        old_tty = None
+        pid = -1
         # Save original terminal settings
-        try:
-            old_tty = termios.tcgetattr(sys.stdin)
-        except BaseException:
-            old_tty = None
-
+        if not non_interactive_mode:
```
zastrowm, line 201: Do we expect this to be a global/runtime property that will never change, or would there be a use case where sometimes you have interactive sometimes not?

I'm wondering if this should be something that should be shared among more tools. Same for user input - when you don't have an interactive mode does user input work?

src/strands_tools/shell.py
zastrowm, line 1: How would we test this in both modes, out of curiosity?  Could we document the steps so that in the future we know how to verify it?

tests/test_shell.py
JackYPCOnline: Yes, I also add integration test for non_interactive_mode=True

src/strands_tools/shell.py
JackYPCOnline: Returning the differnt exit code from the underlying process is crucial for debugging.

src/strands_tools/shell.py
JackYPCOnline: 1.I think a global runtime property makes more sense. This creates a clear, predictable contract for all tools.
2. You're right, user input cannot work in non-interactive mode. Any tool requiring input will now check for the global flag first and fail immediately if it's set, preventing hanging processes.

src/strands_tools/shell.py
JackYPCOnline: 1. Testing Non-Interactive Mode
We achieve this by setting the global flag and asserting the tool succeeds without asking for input.
2.  Testing Interactive Mode 
I could  simulate a user confirming the action by patching get_user_input with Y / N.

src/strands_tools/shell.py
JackYPCOnline: Also, BYPASS_TOOL_CONSENT works as ideal, when non_interactive_mode = false and BYPASS_TOOL_CONSENT = true, user don't need to input.
But somehow, BYPASS_TOOL_CONSENT does a smiliar job but standardizing on something like "STRANDS_NON_INTERACTIVE" might be a better option,

zastrowm: Yeah - perhaps switching to STRANDS_NON_INTERACTIVE for now makes more sense?  We can re-evaluate in the future if we want it to be automatic

src/strands_tools/shell.py
```diff
@@ -184,17 +191,18 @@ def __init__(self, timeout: int = None) -> None:
         self.exit_code = None
         self.error = None
 
-    def execute_with_pty(self, command: str, cwd: str) -> Tuple[int, str, str]:
+    def execute_with_pty(self, command: str, cwd: str, non_interactive_mode: bool) -> Tuple[int, str, str]:
```
zastrowm, line 194: Could we default `non_interactive_mode` to `false` to avoid always needing to pass it in?  That would clean up the tests/caller when using this explicitly?

PR #88, opened 2025-06-16
feat(tools): add use_browser tool to Strands tools repository

jimbrub:
## Description
Adding a use_browser tool, that gives an agent actions, so they interact with browsers (just chromium right now). This tool relies on Playwright to perform automated actions on a browser. Playwright allows for a browser to be launched with different launch options. One of these launch options is persistent_context, which allows users to launch a browser and load in the saved data from whatever the specified user_data_dir is. This feature is used to save cookie settings, local storage, sessions storage, browser history, saved passwords, etc. Right now you can ask an agent to use the use_browser tool to launch a browser with persistent_context set to true and specify a file path for user_data_dir. Then, any usernames, bookmarks you create, or search history should be stored in the user_data_dir and then the browser can be relaunched at a later date. 

## Related Issues
(https://github.com/strands-agents/private-sdk-python-staging/issues/72)

## Documentation PR
Will need to add documentation about use_browser tool. 

## Type of Change
- [x] New Tool

## Testing

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

## Checklist


src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+            default_launch_options = {"headless": False, "args": ["--window-size=1280,800"]}
+
+            if launch_options:
+                default_launch_options.update(launch_options)
+
+            # Handle persistent context
+            if launch_options and launch_options.get("persistent_context"):
+                user_data_dir = launch_options.get("user_data_dir")
+                if user_data_dir:
+                    logger.debug(f"Creating persistent context with user_data_dir: {user_data_dir}")
```
mehtarac: Imo, let's be conservative on the information we are logging. This log is a good example of being useful during development but when we are shipping it, we don't want so much information on the screen that it's hard to keep track of. There are other logging statements that I believe can be removed and I commented on them

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+        if cleanup_errors:
+            for error in cleanup_errors:
+                logger.warning(error)
+        else:
+            logger.info("Cleanup completed successfully")
+
+    async def handle_action(self, action: str, **kwargs) -> List[Dict[str, str]]:
+        """Handle both high-level actions and direct CDP commands."""
+        logger.debug(f"Handling action: {action}")
+        logger.debug(f"Action arguments: {kwargs}")
```
mehtarac: Can remove the kwargs log statement

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+            page, cdp = await self.ensure_browser(
+                launch_options=launch_options,
+            )
+
+            # High-level actions
+            if action == "connect":
+                result = await self._handle_connect_action(launch_options)
+
+            elif action == "navigate":
+                logger.info("attempting navigate")
```
mehtarac: can remove this since none of the other actions have a log

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+                wait_time = kwargs["wait_for"]
+                logger.debug(f"Waiting for {wait_time}ms")
+                await page.wait_for_timeout(wait_time)
+
+            logger.debug(f"Action '{action}' completed successfully")
+            return result
+        except Exception as e:
+            logger.error(f"Error executing action '{action}': {str(e)}")
+            if "browser has been closed" in str(e) or "browser disconnected" in str(e):
+                logger.debug("Cleaning up browser due to error or non-persistent session")
```
mehtarac: Let's make this an error level log since it's under the Exception condition

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+
+    # Cookie management actions
+    async def _handle_get_cookies_action(self):
+        logger.debug("Getting cookies")
+        cookies = await self._context.cookies()
+        return [{"text": f"Cookies: {json.dumps(cookies, indent=2)}"}]
+
+    async def _handle_set_cookies_action(self, args):
+        cookies = args.get("cookies", [])
+        logger.debug(f"Setting cookies: {cookies}")
```
mehtarac: Can we remove this log imo.

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+            {"action": "click", "args": {"selector": "#submit"}}
+        ])
+    """
+
+    logger.info(f"use_browser tool called with action: {action}")
+
+    if actions:
+        logger.info(
+            f"Multiple actions requested: {[a.get('action') for a in actions if isinstance(a, dict) and 'action' in a]}"
+        )
```
mehtarac: should this be in a condition where if there is one action we log one action else we log all the actions. The way it is currently, I believe it'll log one action and also multiple actions if there are.

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+            cancellation_reason = (
+                user_input if user_input.strip() != "n" else get_user_input("Please provide a reason for cancellation:")
+            )
+            error_message = f"Python code execution cancelled by the user. Reason: {cancellation_reason}"
+            return {
+                "status": "error",
+                "content": [{"text": error_message}],
+            }
+
+    logger.info(f"Tool parameters: {locals()}")
```
mehtarac: Curious to understand if this is useful?

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+                    args=action_args,
+                    selector=action_selector,
+                    wait_for=action_wait_for,
+                )
+                results.extend(content)
+            return results
+
+        # Run all actions in a single event loop call
+        all_content = _playwright_manager._loop.run_until_complete(run_all_actions())
+        logger.debug(f"Results from run_until_complete: {all_content}")
```
mehtarac: Could we remove this log statement. If we think there is something specific that is useful from the all_content, then let's log that instead. (ex: result status)

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+        return "\n".join([item["text"] for item in all_content])
+
+    except Exception as e:
+        logger.error(f"Error in use_browser: {str(e)}")
+        logger.info("Cleaning up browser due to explicit request or error with non-persistent session")
```
mehtarac: Let's make this an error level log

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+        return [{"text": f"Error: {param_name} required for {action_name}"}]
+    return None
+
+
+@tool
+def use_browser(
+    url: str = None,
+    wait_time: int = 1,
+    action: str = None,
+    new_tab: bool = False,
```
mehtarac: for variables with default values, let's follow the convention of them being configurable by env variables. Mainly only for variables where the default value is not None. Other tools currently do this. ([ref](https://github.com/strands-agents/tools/blob/main/src/strands_tools/current_time.py#L37))

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+
+            # Handle wait_for if specified
+            if kwargs.get("wait_for"):
+                wait_time = kwargs["wait_for"]
+                logger.debug(f"Waiting for {wait_time}ms")
+                await page.wait_for_timeout(wait_time)
+
+            logger.debug(f"Action '{action}' completed successfully")
+            return result
+        except Exception as e:
```
mehtarac: can we add an error to check for `ERR_SOCKET_NOT_CONNECTED`. While testing I noticed I was running into this. Maybe something like:

```
if "ERR_SOCKET_NOT_CONNECTED" in str(e):
            return [{"text": f"Error: Connection issue detected. Please verify network connectivity and try again."}]
```

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,566 @@
...
+            logger.error(f"Error executing action '{action}': {str(e)}")
```
mehtarac: As important as it is to catch these errors, I think it'll be a valuable mechanism to add a retry logic so that if an action fails due to transient errors, it can pass upon retrying a max number of times (suggest 3 but happy to hear alternate thoughts). I am thinking we can perhaps make a new function. Some pseudocode:


```
   async def retry_action(self, operation, max_retries=3, delay=1.0):
       """Retry an async operation"""
       last_exception = None
       for attempt in range(max_retries):
           try:
               return await action()
           except Exception as e:
               last_exception = e
               if attempt < max_retries - 1:
                   wait_time = delay * (2 ** attempt)
                   logger.warning(f"Action failed, retrying in {wait_time}s: {str(e)}")
                   await asyncio.sleep(wait_time)
       
       logger.error(f"Action failed after {max_retries} attempts: {str(last_exception)}")
       raise last_exception
   ```

File: src/strands_tools/use_browser.py [hunk 13]
```diff
@@ -0,0 +1,566 @@
...
+        text = args.get("text")
+        error = validate_required_param(selector, "selector", "type")
+        if error:
+            return error
+        error = validate_required_param(text, "text", "type")
+        if error:
+            return error
+        await page.fill(selector, text)
+        return [{"text": f"Typed '{text}' into {selector}"}]
+
```
mehtarac: This is a higher level design comment and more aiming to say this to start a discussion so feel free to disagree. I know I am the one who initially suggested to make all the different actions into their own methods. I was primarily motivated to suggest this because different actions have different parameters and I thought it would be a more organized way to keep track of everything. However, I realize that this is introducing a lot of code duplication and redundancy. 

- Currently there is a long list of `if/elif` statements in the `handle_action()` function and then each of those conditions call a dedicated function for that action. 

I believe can consolidate this by creating a generic action handler method to reduce repetition of `_handle_*_action` methods. 

For example:

```
# Define action configurations - maps action names to their requirements and methods
    action_configs = {
        "navigate": {
            "method": page.goto(),
            "required_params": [("url", True, str)],
            "result_template": "Navigated to {url}"
        },
   }
   
   ```
   
   ```
   def _handle_action():
       # Get the configuration for this action
    config = action_configs.get(action_name)
    if not config:
        return [{"text": f"Error: Unknown action '{action_name}'"}]
     
     # Validate required parameters
    required_params = config.get("required_params", [])
    if required_params:
        validation_error = validate_required_params(args, required_params, action_name)
        if validation_error:
            return validation_error
 ```

Comment (jimbrub) [reply src/strands_tools/use_browser.py]: I like this way of doing it, I think it would reduce the amount of code and make it a bit more readable, and easier to understand if actions needed to be added.

Comment (jimbrub) [reply src/strands_tools/use_browser.py]: I like the suggestion you are making here. I think it would reduce the amount of code and make it a bit more readable, and easier to understand if actions needed to be added.

File: src/strands_tools/use_browser.py (hunk 9 above)
Comment (jimbrub) [reply src/strands_tools/use_browser.py]: makes sense, i will also clean up the other logging. A lot of the logging was for debugging

File: src/strands_tools/use_browser.py (hunk 7 above)
Comment (jimbrub) [reply src/strands_tools/use_browser.py]: This logs the variables that the tool is being called with: i.e. `INFO | use_browser | Tool parameters: {'url': 'https://www.google.com', 'wait_time': 1, 'action': 'navigate', 'new_tab': False, 'selector': None, 'input_text': None, 'script': None, 'cdp_method': None, 'cdp_params': None, 'launch_options': None, 'actions': None, 'key': None, 'strands_dev': True}`

File: src/strands_tools/use_browser.py (hunk 6 above)
Comment (jimbrub) [reply src/strands_tools/use_browser.py]: Ah i see now. I added that logging when debugging. I think I will remove those lines.

File: src/strands_tools/use_browser.py (hunk 11 above)
Comment (jimbrub) [reply src/strands_tools/use_browser.py]: I can add a check for that. Just curious, how was this error occurring?

File: src/strands_tools/use_browser.py (hunk 12 above)
Comment (jimbrub) [reply src/strands_tools/use_browser.py]: I agree. There are some times that navigate "fails" when the agent just needed to wait longer, so this could save tokens

File: src/strands_tools/use_browser.py (hunk 13 above)
Comment (jimbrub) [reply src/strands_tools/use_browser.py]: Some of the actions might be too specific like the "close" action, but i think action_configs could be incorporated in the actions that accept the same argurments.

File: src/strands_tools/use_browser.py [hunk 14]
```diff
@@ -32,6 +42,9 @@
 # Apply nested event loop support
 nest_asyncio.apply()
 
+# Environment Variables
+default_wait_time = int(os.getenv("DEFAULT_WAIT_TIME", 1))
```
mehtarac: Let's add all the the environment variables that this tool has in the README file in the Environment Variables section similar to other tools.

src/strands_tools/use_browser.py
```diff
@@ -43,53 +56,189 @@ def __init__(self):
         self._cdp_client = None
         self._user_data_dir = None
         self._profile_name = None
+        self._tabs = {}  # Dictionary to track tabs by ID
+        self._active_tab_id = None  # Currently active tab ID
         self._loop = asyncio.new_event_loop()
         asyncio.set_event_loop(self._loop)
+        self.action_configs = {
+            "navigate": {
+                "method": lambda page, args: page.goto(args["url"]),
```
mehtarac: I think we should validate the url here. In my local testing I had a query to go to an invalid URL such as "https://.com" or "invalid-url-with-no-protocol", etc. and the great news is that the rery mechanism worked! (The tool retried 3 times), however, it threw an error `ERR_NAME_NOT_RESOLVED` error. I think we can have a basic check whether a url is valid or not and/or catch this specific error so we can have a helpful message for the user to check the url in their query. 

The test script i used is:

```

import asyncio
from strands import Agent
from strands_tools.use_browser_new import use_browser_new

def main():
    # Create an agent with the tool
    agent = Agent(tools=[use_browser_new])
    
    print("Testing invalid URL handling...")
    
    # Test with completely invalid URL format
    print("\nTest 1: Completely invalid URL format")
    response = agent("navigate to invalid-url-with-no-protocol")
    print(f"Response: {response}")
    
    # Test with invalid domain
    print("\nTest 2: Invalid domain")
    response = agent("navigate to https://thisisnotarealdomainxyz123456.com")
    print(f"Response: {response}")
    
    # Test with malformed URL
    print("\nTest 3: Malformed URL")
    response = agent("navigate to http://.com")
    print(f"Response: {response}")

if __name__ == "__main__":
    main()
```

src/strands_tools/use_browser.py
```diff
@@ -43,53 +56,189 @@ def __init__(self):
...
         logger.debug("Ensuring browser is running...")
 
-        if self._playwright is None:
-            self._playwright = await async_playwright().start()
+        try:
+            if self._playwright is None:
+                self._playwright = await async_playwright().start()
 
-            default_launch_options = {"headless": False, "args": ["--window-size=1280,800"]}
+                default_launch_options = {"headless": False, "args": ["--window-size=1280,800"]}
```
mehtarac: I think we can allow these to be configurable by environment variables. For example:

os.getenv("BROWSER_HEADLESS", "false").lower() == "true"
int(os.getenv("BROWSER_WIDTH", "1280"))
int(os.getenv("BROWSER_HEIGHT", "800"))

src/strands_tools/use_browser.py
```diff
@@ -110,260 +259,348 @@ async def cleanup(self):
...
+            logger.error(f"Error executing action '{action}': {str(e)}")
+            if "ERR_SOCKET_NOT_CONNECTED" in str(e):  # Adding special case for when network connection issues
+                return [{"text": "Error: Connection issue detected. Please verify network connectivity and try again."}]
+            if "browser has been closed" in str(e) or "browser disconnected" in str(e):
+                await self.cleanup()
+            return [{"text": f"Error: {str(e)}"}]
 
-            elif action == "click":
-                result += await self._handle_click_action(page, args)
+    async def retry_action(self, action_func, max_retries=3, delay=1.0, action_name=None, args=None):
```
mehtarac: Let's make this configurable by environment variables. Along with the `delay` parameter.

src/strands_tools/use_browser.py
```diff
@@ -43,53 +56,189 @@ def __init__(self):
...
-        return self._page, self._cdp_client
+            if not self._page:
+                raise ValueError("Browser initialized but page is not available")
+
+            return self._page, self._cdp_client
+
+        except Exception as e:
+            logger.error(f"Failed to initialize browser: {str(e)}")
+            # Clean up any partial initialization
+            await self.cleanup()
```
mehtarac, line 332: Higher level comment/question regarding error. Do you think we should handle the `WebError` specifically? [Ref](https://github.com/microsoft/playwright-python/blob/0ff7fc983ef39ed6121ffbac4eaf15e9b02a6aba/playwright/_impl/_web_error.py#L22).

```
A Playwright WebError represents an unhandled exception that occurs within a web page being controlled by Playwright.
```
https://playwright.dev/docs/api/class-weberror


Similary should we handle `TimeOutError` https://playwright.dev/docs/api/class-timeouterror

Happy to hear your thoughts if you disagree or if you think it's already being handled

src/strands_tools/use_browser.py
```diff
@@ -43,53 +56,189 @@ def __init__(self):
...
+                "required_params": [],
+                "post_action": lambda page: page.wait_for_load_state("networkidle"),
+                "result_template": "Navigated back",
+            },
+            "forward": {
+                "method": lambda page, args: page.go_forward(),
+                "required_params": [],
+                "post_action": lambda page: page.wait_for_load_state("networkidle"),
+                "result_template": "Navigated forward",
+            },
```
mehtarac: looks like `args` parameter is not used here, is it needed?

src/strands_tools/use_browser.py
```diff
@@ -43,53 +56,189 @@ def __init__(self):
...
+            "switch_tab": {
+                "method": lambda page, args: self._switch_to_tab(args.get("tab_id")),
+                "required_params": [("tab_id", str)],
+                "result_template": "Switched to tab: {tab_id}",
+            },
+            "close_tab": {
+                "method": lambda page, args: self._close_tab_by_id(args.get("tab_id", self._active_tab_id)),
+                "required_params": [],
+                "result_template": "Tab closed successfully",
+            },
```
mehtarac: Similarly, the `page` parameter is not used in these. Curious if they can be removed

tests/test_use_browser.py
```diff
@@ -0,0 +1,239 @@
...
+@pytest.mark.parametrize("error_scenario", ERROR_SCENARIOS)
+def test_complex_error_conditions(setup_test_environment, mock_browser_manager, error_scenario):
+    action, args, expected_error = error_scenario
+    mock_browser_manager._loop.run_until_complete.return_value = [{"text": expected_error}]
+    result = use_browser(action=action, **args)
+    assert_browser_action(result, expected_error)
+
+
+@pytest.mark.parametrize("launch_options", LAUNCH_OPTIONS_SCENARIOS)
+def test_launch_options_combinations(setup_test_environment, mock_browser_manager, launch_options):
```
mehtarac: nit: doesn't look like `setup_test_environment` parameter is used here. Can probably remove it

mehtarac: Similarly let's check if there are other tests with parameters that are unused and can be removed.

src/strands_tools/use_browser.py
mehtarac: nit: maybe this can be a debug level log instead of info

src/strands_tools/use_browser.py
mehtarac: Was running into this error when chromium was having issues opening

src/strands_tools/use_browser.py
jimbrub: Oooo nice nice. We can also make the paths for the screenshots and for the user_data_dir environmental variables

src/strands_tools/use_browser.py
jimbrub: Having a mechanism to see if a url exists or not would be useful, because it takes the agent a second to realize when it has gone to a url that does not exist and sometimes it doesnt even realize. For example, if it tries to go to a url that does not exist on GitHub, then it just thinks it needs to sign in and does not see the 404 error

src/strands_tools/use_browser.py
jimbrub: I like better error handling, it will provide more context to the user about the specific error, so they can adjust their prompt/approach accordingly. I see that an all for one error statement might not be the best approach

src/strands_tools/use_browser.py
jimbrub: Ahh, it looks like it is not needed in these configs

src/strands_tools/use_browser.py
jimbrub: I think page might need to be passed, if we are using self, but I am not too sure, I will check

jimbrub:
> High level comment:
> 
> Let's add this tool to the README as well

tests/test_use_browser.py
jimbrub: When I remove the setup_test_environment from there then it does not call the set_test_enviornment fixture when running that test. The fixture sets up a test environment to deal with BYPASS_TOOL_CONSENT. I am not too sure of another way of doing this

zastrowm: Update the fixture to mark it as `autouse`:

```
@pytest.fixture(autouse=True)
```

tests/test_use_browser.py
```diff
@@ -0,0 +1,239 @@
...
+    else:
+        assert expected_text in result
+
+
+# Fixtures
+@pytest.fixture
+def setup_test_environment():
+    """Fixture to set up common test environment"""
+    original_value = os.environ.get("BYPASS_TOOL_CONSENT", None)
+    os.environ["BYPASS_TOOL_CONSENT"] = "true"
```
zastrowm: Using mocking of the environment. See https://github.com/strands-agents/tools/blob/44f5d23370f1350fef90fc0f4f1df7071b9eb3b0/tests/test_environment.py#L29 for example

src/strands_tools/use_browser.py
jimbrub: In the generic_action_handler helper function we are calling each action with the arguments args and pages, as seen here (https://github.com/strands-agents/tools/pull/88/files#diff-6da2a27bbf2ae4dc58907410c58579038912e61a50c91f8947e7585582a65fceR473). I have tried a couple of ways to check the configs and see what arguments it should send, but I haven't been able to get them to work, without pretty much having to revert back to the way before of defining these actions (all with different helper functions). I am open to any ideas on how to resolve this, or if just leave it as is

tests/test_use_browser.py
jimbrub: I see, thank you for this. Autouse works well for the setup for the tests

pyproject.toml
```diff
@@ -38,6 +38,8 @@ dependencies = [
     "tenacity>=9.1.2,<10.0.0",
     "watchdog>=6.0.0,<7.0.0",
     "slack_bolt>=1.23.0,<2.0.0",
+    "nest-asyncio>=1.5.0,<2.0.0",
+    "playwright>=1.42.0,<2.0.0",
```
zastrowm, line 42: How heavy is this dependency?

I'm leaning towards this being an optional dependency as I don't want everyone to need to download playwright just to use `memory` for instance

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,1010 @@
...
+)
+from rich.console import Console
+from rich.panel import Panel
+from rich.text import Text
+from strands import tool
+
+from strands_tools.utils.user_input import get_user_input
+
+# Only configure this module's logger, not the root logger
+enable_debug = os.getenv("ENABLE_DEBUG_BROWSER_LOGS", "false").lower() == "true"
```
zastrowm: What's our naming convention? I think use a `STRANDS` prefix elsewhere and I wonder if we should add that prefix

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,1010 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG if enable_debug else logging.INFO)
```
zastrowm: This seems like a client-level concern, not a tool level concern.  Any reason we shouldn't let the application/caller set up logging properly?

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,1010 @@
...
+    logger.addHandler(handler)
+logger.propagate = False
+
+console = Console()
+
+# Global browser manager instance
+_playwright_manager = None
+
+# Apply nested event loop support
+nest_asyncio.apply()
```
zastrowm: Can this be done lazily at all?  I'm a little concerned that simply importing the tool changes the behavior

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,1010 @@
...
+if not logger.handlers:
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s"))
+    logger.addHandler(handler)
+logger.propagate = False
+
+console = Console()
+
+# Global browser manager instance
+_playwright_manager = None
```
zastrowm, line 34: Is there a specific reason to have a global instance?  If so, can that reason be added as a comment?

Are there any limitations around multiple agents interacting with this tool?

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,1010 @@
...
+        self._page: Optional[Page] = None
+        self._cdp_client = None
+        self._user_data_dir = None
+        self._profile_name = None
+        self._tabs = {}  # Dictionary to track tabs by ID
+        self._active_tab_id = None  # Currently active tab ID
+        self._loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self._loop)
+        self.action_configs = {
+            "navigate": {
```
zastrowm: Rather than config as code, would it make sense to go to a defined object?

```
class ApiMethods:
    def navigate(page: Page, url: str):
        page.wait_for_load_state("networkidle"))
        return "Navigated to {url}"
    ....
```

I think you'd be able to use reflection at runtime to build up the same data here, but it'd be a more declarative way of exposing actions

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,1010 @@
...
+                "method": lambda page, key: page.keyboard.press(key),
+                "required_args": ["page", "key"],
+                "required_params": [("key", str)],
+                "result_template": "Pressed key: {key}",
+            },
+            "get_text": {
+                "method": lambda page, selector: page.text_content(selector),
+                "required_args": ["page", "selector"],
+                "required_params": [("selector", str)],
+                "post_process": lambda result: result,
```
zastrowm: Is this actually needed?

src/strands_tools/use_browser.py
```diff
@@ -0,0 +1,1010 @@
...
+                "result_template": "Browser closed",
+            },
+        }
+
+    async def ensure_browser(self, launch_options=None, context_options=None):
+        """Initialize browser if not already running."""
+        logger.debug("Ensuring browser is running...")
+
+        # Ensure required directories exist
+        user_data_dir = os.getenv("BROWSER_USER_DATA_DIR", os.path.join(os.path.expanduser("~"), ".browser_automation"))
```
zastrowm: Related to my earlier question - what's our convention with naming?  Should these be prefixed with `STRANDS_`?

Are these parameters only required because we don't provide a way to configure tools otherwise?  What if you implemented it such as:

```
class BrowserTool:
    def __init__(... params...):


    @tool
    def use_browser(...)
           ... 
```


Usage would then be:

```
use_browser = BrowserTool(... params ...).use_browser
```

which would also allow you to isolate state.

(this may or may not work today; but after strands-agents/sdk-python#258 it should work).  We may also want to ping the team to get thoughts on this class based approach for this case

pyproject.toml
jimbrub: I wasn't too sure on the acceptable way to add these dependencies. As long as adding them as optional dependencies keeps the same functionality with the tool, I would rather do it that way.

src/strands_tools/use_browser.py
jimbrub: Would a better convention for this variable be "STRANDS_USE_BROWSER_ENABLE_DEBUG_LOGS"?

src/strands_tools/use_browser.py
jimbrub: I was having an issue with the logs in the past printing out everything, I will re-look at this, logs should be configured on client-level, that makes sense.

src/strands_tools/use_browser.py
jimbrub: There is no reason that I have found to why it needs a global instance, let me test the tool with that line in the tool function

src/strands_tools/use_browser.py
jimbrub: I think I can move this into the initialization code for the BrowserManager class

src/strands_tools/use_browser.py
jimbrub: Ahh, I see let me test it without that and see

src/strands_tools/use_browser.py
mehtarac: I like this approach as well. This way we get access to all the actions of the class in one go.

src/strands_tools/use_browser.py
mehtarac: +1 to Mackenzie's point. It's a small thing but it'll be good to follow the naming convention that we have throughout the repo.

src/strands_tools/use_browser.py
jimbrub: I do like the idea of a class based approach here

pyproject.toml
zastrowm: Yeah, you can make it optional and then the installs are done via:

```
pip install strands-agents-tools[use-browser]
```

Issue #75, opened 2025-06-07
[BUG] generate_image tool not working (required key prompt not found)


haritsE:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

0.1.6

### Tools Package Version

0.1.4

### Tools used

generate_image

### Python Version

3.13.4

### Operating System

macOS 15.4

### Installation Method

pip

### Steps to Reproduce

Install strands with pip
Run the following command:
```
        response = self.agent.tool.generate_image(
            prompt="A futuristic city with flying cars",
            model_id="stability.stable-image-core-v1:1"
        )
        print(response)
```
Got error:
```
{'toolUseId': 'tooluse_generate_image_264683723', 'status': 'error', 'content': [{'text': 'Error generating image: An error occurred (ValidationException) when calling the InvokeModel operation: Malformed input request: #: required key [prompt] not found, please reformat your input and try again.'}]}
```

### Expected Behavior

Image to be generator without validation exception, since prompt param is supplied

### Actual Behavior

Validation error

### Additional Context

Was using the command listed both in the README and in https://github.com/strands-agents/tools/blob/main/src/strands_tools/generate_image.py, but didn't work

pgrayy:
Would you be able to upgrade to the latest version of `strands-agents` and `strands-agents-tools`. The following now works:
```Python
from strands import Agent
from strands_tools import generate_image

agent = Agent(tools=[generate_image])
response = agent.tool.generate_image(prompt="A futuristic city with flying cars", model_id="stability.stable-image-core-v1:1")
```

The generated image was saved locally to `output/a_futurisitc_city_with_flying.png`.

yonib05:
Fixed by #89, which reworked model support in generate_image. The repro in this issue works on current strands-agents-tools (confirmed above). Closing.

PR #89, opened 2025-06-17
fix(generate_image): updating generate_image tool to support additional models in Amazon Bedrock

cerashdan:
## Description

The model currently used in generate_image tool has a hard-coded region for us-west-2. The model is no longer available in that region. The model is also being phased out of Amazon Bedrock. 

This PR will focus on the generate image tool to include the latest supported image generation models and their parameters. The fix has implemented support for all Stable Diffusion and Nova Canvas models. 

## Related Issues

This will fix issue #75 

## Documentation PR

This will ensure that PR for multi modal examples is operating as expected- strands-agents/docs#93

## Type of Change
- [x] Bug fix
- [ ] New Tool
- [ ] Breaking change

## Testing

1. Unit tests, formatter and linters has passed. 
2. Generated 4 invocation to models in scope - generating images using (Nova Canvas, SD Ultra, SD Core and SD 3.5).

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

## Checklist

### DIFF
--- a/src/strands_tools/generate_image.py
+++ b/src/strands_tools/generate_image.py
@@ -95,6 +95,10 @@ TOOL_SPEC = {
                     "description": "Model id for image model, stability.sd3-5-large-v1:0, \
                     stability.stable-image-core-v1:1,  stability.stable-image-ultra-v1:1, or amazon.nova-canvas-v1:0",
                 },
+                "region": {
+                    "type": "string",
+                    "description": "AWS region for the image generation model (default: us-west-2)",
+                },
                 "seed": {
                     "type": "integer",
                     "description": "Optional: Seed for random number generation (default: random)",
@@ -130,10 +134,6 @@ TOOL_SPEC = {
                 "cfg_scale": {
                     "type": "number",
                     "description": "Optional: CFG scale for Nova Canvas model (default: 8.0)",
-                },
-                "number_of_images": {
-                    "type": "integer",
-                    "description": "Optional: Number of images to generate for Nova Canvas model (default: 1)",
                 },
             },
             "required": ["prompt"],
--- a/src/strands_tools/generate_image.py
+++ b/src/strands_tools/generate_image.py
@@ -170,7 +170,7 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
         steps = tool_input.get("steps", 30)
 
         # Create a Bedrock Runtime client
-        client = boto3.client("bedrock-runtime", region_name="us-west-2")
+        client = boto3.client("bedrock-runtime", region_name="us-east-1")
 
         # Format the request payload
         native_request = {
--- a/src/strands_tools/generate_image.py
+++ b/src/strands_tools/generate_image.py
@@ -155,7 +155,6 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
     --------------------
     - prompt: The textual description of the desired image
     - model_id: Specific model to use (defaults to stability.stable-image-core-v1:1)
-
     For Stable Diffusion models:
     - seed: Controls randomness for reproducible results
     - aspect_ratio: Controls the aspect ratio of the generated image
--- a/src/strands_tools/generate_image.py
+++ b/src/strands_tools/generate_image.py
@@ -1,20 +1,19 @@
 """
-Image generation tool for Strands Agent using Stable Diffusion and Nova Canvas models.
+Image generation tool for Strands Agent using Stable Diffusion.
 
 This module provides functionality to generate high-quality images using Amazon Bedrock's
-image generation models based on text prompts. It handles the entire image generation
+Stable Diffusion models based on text prompts. It handles the entire image generation
 process including API integration, parameter management, response processing, and
 local storage of results.
 
 Key Features:
 
 1. Image Generation:
-   • Text-to-image conversion using multiple model providers
+   • Text-to-image conversion using Stable Diffusion models
    • Support for the following models:
         • stability.sd3-5-large-v1:0
         • stability.stable-image-core-v1:1
         • stability.stable-image-ultra-v1:1
-        • amazon.nova-canvas-v1:0
    • Customizable generation parameters (seed, aspect_ratio, output_format, negative_prompt)
 
 2. Output Management:
@@ -48,13 +47,13 @@ agent.tool.generate_image(
     negative_prompt="bad lighting, harsh lighting, abstract, surreal, twisted, multiple levels",
 )
 
-# Using Nova Canvas
+# Using another Stable Diffusion model
 agent.tool.generate_image(
     prompt="A photograph of a cup of coffee from the side",
-    model_id="amazon.nova-canvas-v1:0",
-    width=1024,
-    height=1024,
-    quality="premium",
+    model_id="stability.stable-image-ultra-v1:1",
+    aspect_ratio="1:1",
+    output_format="png",
+    negative_prompt="blurry, distorted",
 )
 ```
 
@@ -76,12 +75,11 @@ STABLE_DIFFUSION_MODEL_ID = [
     "stability.stable-image-core-v1:1",
     "stability.stable-image-ultra-v1:1",
 ]
-NOVA_CANVAS_MODEL_ID = "amazon.nova-canvas-v1:0"
 
 
 TOOL_SPEC = {
     "name": "generate_image",
-    "description": "Generates an image using Stable Diffusion or Nova Canvas based on a given prompt",
+    "description": "Generates an image using Stable Diffusion models based on a given prompt",
     "inputSchema": {
         "json": {
             "type": "object",
@@ -93,7 +91,7 @@ TOOL_SPEC = {
                 "model_id": {
                     "type": "string",
                     "description": "Model id for image model, stability.sd3-5-large-v1:0, \
-                    stability.stable-image-core-v1:1,  stability.stable-image-ultra-v1:1, or amazon.nova-canvas-v1:0",
+                    stability.stable-image-core-v1:1, or stability.stable-image-ultra-v1:1",
                 },
                 "region": {
                     "type": "string",
@@ -118,23 +116,6 @@ TOOL_SPEC = {
                     "description": "Optional: Keywords of what you do not wish to see in the output image. \
                      Max: 10.000 characters.",
                 },
-                "width": {
-                    "type": "integer",
-                    "description": "Optional: Width of the generated image for Nova Canvas model (default: 1024)",
-                },
-                "height": {
-                    "type": "integer",
-                    "description": "Optional: Height of the generated image for Nova Canvas model (default: 1024)",
-                },
-                "quality": {
-                    "type": "string",
-                    "description": "Optional: Quality setting for Nova Canvas model. Options: 'standard' or 'premium' \
-                    (default: 'standard')",
-                },
-                "cfg_scale": {
-                    "type": "number",
-                    "description": "Optional: CFG scale for Nova Canvas model (default: 8.0)",
-                },
             },
             "required": ["prompt"],
         }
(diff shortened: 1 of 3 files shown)


src/strands_tools/generate_image.py
```diff
@@ -170,7 +170,7 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
         steps = tool_input.get("steps", 30)
 
         # Create a Bedrock Runtime client
-        client = boto3.client("bedrock-runtime", region_name="us-west-2")
+        client = boto3.client("bedrock-runtime", region_name="us-east-1")
```
zastrowm: We talked offline; we think this makes more sense to:

 - Update to be a region parameter (maybe defaulting to us-east-1)
 - Update the model-id and body to support additional models that aren't supported right now

cerashdan:
The fix has been implemented and the current update to the branch supports all latest Stable Diffusion models (SD Ultra, SD Core and 3.5) and Nova Canvas. 

Tests has been updated accordingly. 

The current implementation defaults to SD Core which is the successor of SDXL that was phased out. 

@zastrowm could you please review and merge the PR?

src/strands_tools/generate_image.py
```diff
@@ -98,19 +142,28 @@
 }
 
 
+# Create a filename based on the prompt
+def create_filename(prompt: str) -> str:
```
zastrowm, line 128: What was the file name before?

src/strands_tools/generate_image.py
```diff
@@ -161,74 +222,131 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
         tool_use_id = tool["toolUseId"]
         tool_input = tool["input"]
 
-        # Extract input parameters
+        # Extract common input parameters
         prompt = tool_input.get("prompt", "A stylized picture of a cute old steampunk robot.")
-        model_id = tool_input.get("model_id", "stability.stable-diffusion-xl-v1")
+        model_id = tool_input.get("model_id", "stability.stable-image-core-v1:1")
+        region = (
```
zastrowm: If a region is explicitly included, then we should always use that region - otherwise we can smartly default depending on the logic you have here.

(alternatively - would it make more sense to always default to us-east-1)

src/strands_tools/generate_image.py
```diff
@@ -161,74 +222,131 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
...
-                    }
+        client = boto3.client("bedrock-runtime", region_name=region)
+
+        # Initialize variables for later use
+        base64_image_data = None
+        output_format = "jpeg"  # Default format
+
+        # Format the request payload based on model type
+        if (
+            model_id == "stability.sd3-5-large-v1:0"
```
zastrowm: Is it worth updating this to use `if model_id in [...]`

src/strands_tools/generate_image.py
```diff
@@ -161,74 +222,131 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
...
+                              1.stability.sd3-5-large-v1:0 \n \
+                              2. stability.stable-image-core-v1:1 \n \
+                              3. stability.stable-image-ultra-v1:1 \n \
+                              4. amazon.nova-canvas-v1:0"
+                    }
+                ],
+            }
+
+        # If we have image data, process and save it
+        if base64_image_data:
```
zastrowm, line 233: Given the `number_of_images` parameter, this block should be done for each image - otherwise we're just throwing away the results?

src/strands_tools/generate_image.py
```diff
@@ -60,9 +71,17 @@
...
 
+STABLE_DIFFUSION_MODEL_ID = [
+    "stability.sd3-5-large-v1:0",
+    "stability.stable-image-core-v1:1",
+    "stability.stable-image-ultra-v1:1",
+]
+NOVA_CANVAS_MODEL_ID = "amazon.nova-canvas-v1:0"
+
+
 TOOL_SPEC = {
```
zastrowm, line 80: Did you happen to test an agent using this tool?  I'd like to ensure that an agent can successfully identify which parameters are supported for nova vs stability?  If not, then I wonder if these should be two separate methods.

zastrowm (changes requested):

src/strands_tools/generate_image.py
cerashdan: It was the same, I just moved this function up, so that it could be used in the two conditional for Nova or Stable Diffusion models when they generate the file output.

src/strands_tools/generate_image.py
cerashdan: The reason why I did the initialization is that Nova Canvas is available in (us-east-1, ap-northeast-1, eu-west-1) and Stable Diffusion models are only available in (us-west-2).  However, I do understand that this is subject to change as models become available in other regions. I would just make it default to us-west-2 and stable diffusion - image core model.

src/strands_tools/generate_image.py
cerashdan: The main reason for the conditional here is that Nova Canvas has different body request than Stable Diffusion, if we decide to make this tool specific to only Stable Diffusion and create another tool for Nova Canvas - just as Strands has one for [Nova Reels](https://github.com/strands-agents/tools/nova_reels.py). 

Should we make this tool focused on only Stable Diffusion models?

src/strands_tools/generate_image.py
cerashdan: Let us default this to 1 to keep the tool consistent with one image generated per invocation. I have updated.

src/strands_tools/generate_image.py
cerashdan: While testing I realized I did not include region in the tool spec, causing the agent to fail to pass a region parameter even when explicitly mentioned in the input prompt. Here is the test results for both models: 

## Testing with Nova Canvas after adding region id to toolspec: 

**Input prompt:** "generate an image using Nova Canvas of a futuristic city, use aws region us-east-1 and come up with the best parmaeters configuration for Nova canvas, summarize what you selected in the output"

**logging**: 
tool=<{'toolUseId': 'tooluse_Rcxh-oxRRXqAVMB3OjL3-w', 'name': 'generate_image', 'input': {'model_id': 'amazon.nova-canvas-v1:0', 'prompt': 'A breathtaking futuristic city skyline with gleaming skyscrapers, flying vehicles, holographic billboards, elevated transit systems, and lush vertical gardens integrated into the architecture, dramatic lighting, photorealistic, highly detailed', 'region': 'us-east-1', 'width': 1024, 'height': 1024, 'cfg_scale': 8.0, 'quality': 'premium'}}

**Agent output:** 
For this futuristic city image, I configured Nova Canvas with the following parameters:

1. **Model**: amazon.nova-canvas-v1:0
2. **AWS Region**: us-east-1 (as you specifically requested)
3. **Image Size**: 1024 x 1024 pixels (standard square format for balanced detail)
4. **CFG Scale**: 8.0 (standard setting for good prompt adherence without limiting creativity)
5. **Quality**: Premium (selected for maximum image quality and detail)
6. **Prompt**: Detailed description focusing on futuristic cityscape elements including skyscrapers, flying vehicles, holographic billboards, transit systems, and vertical gardens

The result is a stunning futuristic cityscape with excellent detail showing:
- Soaring skyscrapers with vertical gardens integrated into their design
- Flying vehicles/drones navigating between buildings
- Multi-level transit systems and roadways
- Digital billboards and displays
- A harmonious balance between technology and nature
- Beautiful lighting with natural sunlight illuminating the scene


 **Generated image:** 
a_breathtaking_futuristic_city_skyline

## Testing stable diffusion - image core

**Input prompt**: generate an image using Stable Diffusion model image core of a futuristic city, use aws region us-west-2 and come up with the best parmaeters configuration for Stable Diffusion model, summarize what you selected in the output

**logging**: tool=<{'toolUseId': 'tooluse_06xJ62_URCy9DC6HDSGASw', 'name': 'generate_image', 'input': {'model_id': 'stability.stable-image-core-v1:1', 'prompt': 'Futuristic city with sleek skyscrapers, flying vehicles, advanced transportation systems, holographic displays, neon lights, clean energy infrastructure, and a vibrant urban environment', 'region': 'us-west-2', 'aspect_ratio': '16:9'}}>

**Agent output:** 
## Futuristic City Image - Parameter Selection Summary

I've generated an impressive futuristic cityscape using the Stable Image Core model with the following optimized parameters:

### Parameters Selected:
- **Model**: stability.stable-image-core-v1:1
  - This model provides excellent detail and composition for architectural and city scenes
- **Region**: us-west-2 (as requested)
- **Aspect Ratio**: 16:9
  - Selected this widescreen format to capture the expansive cityscape
- **Prompt**: "Futuristic city with sleek skyscrapers, flying vehicles, advanced transportation systems, holographic displays, neon lights, clean energy infrastructure, and a vibrant urban environment"
  - Detailed prompt that specifies key elements of a futuristic metropolitan landscape

futuristic_city_with_sleek_skyscrapers

src/strands_tools/generate_image.py
```diff
@@ -120,11 +154,15 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
     Generation Parameters:
     --------------------
     - prompt: The textual description of the desired image
-    - model_id: Specific model to use (defaults to stable-diffusion-xl-v1)
+    - model_id: Specific model to use (defaults to stability.stable-image-core-v1:1)
+
```
zastrowm:
```suggestion
```

src/strands_tools/generate_image.py
```diff
@@ -120,11 +154,15 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
...
+    For Stable Diffusion models:
```
zastrowm:
```suggestion
```

src/strands_tools/generate_image.py
```diff
@@ -161,74 +196,91 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
...
+        # Extract common input parameters
         prompt = tool_input.get("prompt", "A stylized picture of a cute old steampunk robot.")
-        model_id = tool_input.get("model_id", "stability.stable-diffusion-xl-v1")
+        model_id = tool_input.get("model_id", "stability.stable-image-core-v1:1")
+        region = tool_input.get("region", "us-west-2")
         seed = tool_input.get("seed", random.randint(0, 4294967295))
-        style_preset = tool_input.get("style_preset", "photographic")
-        cfg_scale = tool_input.get("cfg_scale", 10)
-        steps = tool_input.get("steps", 30)
+        negative_prompt = tool_input.get("negative_prompt", "bad lighting, harsh lighting")
```
zastrowm, line 205: Should we call out these default values in the tool description?

src/strands_tools/generate_image.py
```diff
@@ -161,74 +196,91 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
...
-                        "source": {"bytes": base64.b64decode(base64_image_data)},
+        client = boto3.client("bedrock-runtime", region_name=region)
+
+        # Initialize variables for later use
+        base64_image_data = None
+        output_format = "jpeg"  # Default format
+
+        # Format the request payload based on model type
+        # Validate model ID is a supported Stable Diffusion model
+        if model_id in STABLE_DIFFUSION_MODEL_ID:
```
zastrowm: Could we remove this check and just catch the error from bedrock instead?  If a new model is introduced could we just have it "work" without needing to update the tool?

zastrowm (changes requested):

src/strands_tools/generate_image.py
cerashdan: that's a good call out, I have removed it and generated new images with an agent. 

One interesting observation, the model stability.stable-image-core-v1:0, is no longer listed under supported models in [supported models in Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html) however, we can successfully call it when removing the conditional if statement as it has the same body format.

src/strands_tools/generate_image.py
zastrowm: I still think we should get rid of the "For Stable Diffusion models" part given that it only supports stable-diffusion

Issue #17, opened 2025-05-19
[FEATURE] Improve the tools experience on windows


zastrowm:
### Problem Statement

Attempting to import some of the tools on Windows will result in errors during import due to unsupported imports or libraries (example: #15).  To better support consumers on windows it would be beneficial if we documented what support we have on windows and/or gave more explicit error messages.

### Proposed Solution

Multi-point:

- [x] Add GitHub action workflows to validate behavior on windows
- [ ] Properly mock get_user_input in `test_environment`
- [ ] Attempt to make tools work on windows
    - [ ] If not possible, provide a better error message on windows

### Use Case

When reading the documentation or attempting to import tools on windows, it would either work or the user would receive a better error message.

### Alternatives Solutions

It could just be documented that tools are a best effort on windows

### Additional Context

Reported bugs:

 - strands-agents/tools#15
 - strands-agents/sdk-python#53

zastrowm:
With #18 merge we're now running tests against Windows & MacOS.

With that, I know that:

 - shell
 - python_repl
 - environment

are questionable in their level of support

zastrowm:
Status update:
 - We got the agent-builder running on windows by excluding tools that aren't supported 
 - We documented the tools that are known not working on windows

For the foreseeable future we won't do further updates on this until we receive user feedback that this is high priority compared  to other features

cagataycali:
Closing due inactivity

PR #92, opened 2025-06-18
fix: Document which tools are not supported on windows

zastrowm:
## Description

Per #17 document which tools are confirmed as currently not supported on windows

## Related Issues

#17

## Documentation PR

N/A

## Type of Change
- [X] Bug fix

## Testing

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

## Checklist
- [X] I have added tests that prove my fix is effective or my feature works (NA)

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,9 +95,9 @@ Below is a comprehensive table of all available tools, how to use them with an a
 | file_read | `agent.tool.file_read(path="path/to/file.txt")` | Reading configuration files, parsing code files, loading datasets |
 | file_write | `agent.tool.file_write(path="path/to/file.txt", content="file content")` | Writing results to files, creating new files, saving output data |
 | editor | `agent.tool.editor(command="view", path="path/to/file.py")` | Advanced file operations like syntax highlighting, pattern replacement, and multi-file edits |
-| shell | `agent.tool.shell(command="ls -la")` | Executing shell commands, interacting with the operating system, running scripts |
+| shell* | `agent.tool.shell(command="ls -la")` | Executing shell commands, interacting with the operating system, running scripts |
 | http_request | `agent.tool.http_request(method="GET", url="https://api.example.com/data")` | Making API calls, fetching web data, sending data to external services |
-| python_repl | `agent.tool.python_repl(code="import pandas as pd\ndf = pd.read_csv('data.csv')\nprint(df.head())")` | Running Python code snippets, data analysis, executing complex logic with user confirmation for security |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -155,6 +157,8 @@ results = agent.tool.shell(command=["mkdir -p test_dir", "cd test_dir", "touch t
 agent.tool.shell(command="risky-command", ignore_errors=True)
 ```
 
+*Note: `shell` does not work on Windows.*
```
poshinchen: nit: would it be clearer if the note is located after the header ([line 142](https://github.com/strands-agents/tools/pull/92/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R142))?

README.md
```diff
@@ -200,6 +204,8 @@ processed.head()
 """)
 ```
 
+*Note: `python_repl` does not work on Windows.*
```
poshinchen: nit: same as above

README.md
```diff
@@ -114,14 +114,16 @@ Below is a comprehensive table of all available tools, how to use them with an a
 | current_time | `agent.tool.current_time(timezone="US/Pacific")` | Get the current time in ISO 8601 format for a specified timezone |
 | sleep | `agent.tool.sleep(seconds=5)` | Pause execution for the specified number of seconds, interruptible with SIGINT (Ctrl+C) |
 | agent_graph | `agent.tool.agent_graph(agents=["agent1", "agent2"], connections=[{"from": "agent1", "to": "agent2"}])` | Create and visualize agent relationship graphs for complex multi-agent systems |
-| cron | `agent.tool.cron(action="schedule", name="task", schedule="0 * * * *", command="backup.sh")` | Schedule and manage recurring tasks with cron job syntax |
+| cron* | `agent.tool.cron(action="schedule", name="task", schedule="0 * * * *", command="backup.sh")` | Schedule and manage recurring tasks with cron job syntax <br> **Does not work on Windows |
```
Unshure, line 117: Just curious, why is there no cron example like the other two?

zastrowm: Not sure.

If I had to guess - we wouldn't want to bloat the page with examples of every tool so we instead chose a subset of important tools

README.md
zastrowm: It was an intentional choice to put it at the end (so as not to clutter the example), but I was on the fence.

But the fact that you asked make me reconsider and I'll move it

poshinchen (approved):

PR #94, opened 2025-06-19
test(memory): add an end-to-end integration test

JackYPCOnline:
## Description
This RP is the entrypoint for our integration tests. This test creates real AWS resources (IAM Role, OpenSearch Collection, Bedrock KB) to validate the end-to-end functionality of the memory tool. 

Refactor is automated by hatch fmt.

## Related Issues

## Documentation PR

## Type of Change
- [ ] Bug fix
- [ ] New Tool
- [ ] Breaking change
- [ x] Other (please describe): test

## Testing

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

## Checklist
- [ x] I have added tests that prove my fix is effective or my feature works

### DIFF
--- a/tests-integ/test_memory_tool.py
+++ b/tests-integ/test_memory_tool.py
@@ -2,18 +2,7 @@
 Integration test for the Bedrock Knowledge Base (memory) tool.
 
 This test creates real AWS resources (IAM Role, OpenSearch Collection, Bedrock KB)
-to validate the end-to-end functionality of the memory tool. It is designed to
-be resilient and re-use existing resources to speed up local development.
-
-Prerequisites:
-- AWS credentials must be configured in the environment.
-- The AWS identity must have permissions to manage IAM, OpenSearch Serverless,
-  and Bedrock KB resources.
-
-Configuration:
-- STRANDS_TEARDOWN_RESOURCES (env var): Set to "false" to prevent the automatic
-  deletion of AWS resources after the test run. This is useful for speeding
-  up local development by re-using the same Knowledge Base.
+to validate the end-to-end functionality of the memory tool.
 """
 
 import json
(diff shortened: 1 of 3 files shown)


tests-integ/test_memory_tool.py
```diff
@@ -0,0 +1,461 @@
...
+
+logger = logging.getLogger(__name__)
+
+AWS_REGION = "us-east-1"
+# Use a standard embedding model for the KB
+EMBEDDING_MODEL_ARN = f"arn:aws:bedrock:{AWS_REGION}::foundation-model/amazon.titan-embed-text-v1"
+EMBEDDING_DIMENSION = 1536  # Dimension for amazon.titan-embed-text-v1
+
+
+def _get_boto_clients():
```
zastrowm: Any reason to not each one of these a fixture instead of a method?

tests-integ/test_memory_tool.py
```diff
@@ -0,0 +1,461 @@
...
+        created_resources["ds_id"] = ds_res["dataSource"]["dataSourceId"]
+
+        #  Do not need to wait for CUSTOM data source to become ACTIVE
+        time.sleep(10)
+
+        logger.info("All new resources are ready.")
+        yield created_resources["kb_id"]
+
+    finally:
+        if os.environ.get("STRANDS_TEARDOWN_RESOURCES", "true").lower() == "true":
```
zastrowm: Extract `os.environ.get("STRANDS_TEARDOWN_RESOURCES", "true").lower() == "true"` to a variable to the top to make it clear. 

If we end up re-using this in other tests, let's extract to a central location (but no need now)

tests-integ/test_memory_tool.py
```diff
@@ -0,0 +1,461 @@
...
+                    clients["iam"].delete_role(RoleName=resource_names["role_name"])
+                except ClientError:
+                    logger.warning("Could not delete IAM role.")
+            logger.info("Teardown complete.")
+        else:
+            logger.info("Skipping teardown of AWS resources as per STRANDS_TEARDOWN_RESOURCES setting.")
+
+
+@patch.dict(os.environ, {"BYPASS_TOOL_CONSENT": "true"})
+def test_memory_integration_store_and_retrieve(managed_knowledge_base):
```
zastrowm: In pactice, how long as this taken to run?

tests-integ/test_memory_tool.py
```diff
@@ -0,0 +1,461 @@
...
+    for _ in range(12):
+        retrieve_result = agent.tool.memory(
+            action="retrieve",
+            query="The secret password for the test is",
+            STRANDS_KNOWLEDGE_BASE_ID=kb_id,
+            region_name=AWS_REGION,
+            min_score=0.0,
+            max_results=5,
+        )
+        assert retrieve_result["status"] == "success", f"Retrieve failed: {retrieve_result}"
```
zastrowm: Does retrieve succeed even if we get zero results?

tests-integ/test_memory_tool.py
```diff
@@ -0,0 +1,461 @@
...
+
+        full_retrieved_text = " ".join(item.get("text", "") for item in retrieve_result.get("content", []))
```
zastrowm: Do we always get the most recent result first?  I'm just worried that with enough runs the latest one might be pushed past the 5 results item

We could mitigate that by by including the first couple letters of the secret:

```
The secret password for the test is {unique_content[:5]}`
```

(not sure what the exact syntax is)

tests-integ/test_memory_tool.py
```diff
@@ -0,0 +1,461 @@
...
+        except ClientError as e:
+            if "ResourceNotFoundException" not in str(e):
+                raise e
+            logger.info(f"Resource '{resource_name}' not found yet, continuing to wait...")
+        time.sleep(delay_seconds)
+    raise TimeoutError(f"Timed out waiting for resource '{resource_name}' to become '{success_status}'.")
+
+
+@pytest.fixture(scope="module")
+def managed_knowledge_base():
```
zastrowm, line 36: For readability purposes, what would you think (and is it even possible) to put the KB-creation stuff into a class and move the class to the end of the file?

I'm thinking something like:

```
@pytest.fixture(scope="module")
def managed_knowledge_base():
   kb_helper = KbHelper() # improve naming

   existing = kb_helper.try_get_existing()
   if existing is not None:
      yield existing
    else
       new_kb = kb_helper.create()
       yield new_kb
       if SHOULD_DELETE_RESOURCES:
           kb_helper.destroy()
```

Right now it's really hard to tell what the fixture is doing, but I think that'd make it more readable + refocus the file on the test details instead of the KB-creation details.

If we can't move the class to the bottom of the file, maybe a separate file for readability?

tests-integ/test_memory_tool.py
JackYPCOnline: about 2-3 mins

tests-integ/test_memory_tool.py
JackYPCOnline: Make it a function under Kb helper class now.

tests-integ/test_memory_tool.py
JackYPCOnline: Will refactor this one.

tests-integ/test_memory_tool.py
JackYPCOnline: Yes, status may be 'success' even if no result, but final assertion checks that our content is in the results, so test will fail if not retrieved

tests-integ/test_memory_tool.py
JackYPCOnline: Great idea!! The retrieval query is now using uuid to avoid result-ordering issues. I think it will be semantic relevance ordering.

tests-integ/test_memory_tool.py
JackYPCOnline: Refactored fixture and test code to be easier to read.

zastrowm (approved):

PR #96, opened 2025-06-19
Feat: Improve mem0 memory tool to support storing short-term memories

deshraj:
## Description
This PR adds support for storing the short term memories for agents in the `mem0_memory` tool for managing memories (store, retrieve, list, delete, and history) using Mem0 as the backend. 

## Type of Change
- [ ] Bug fix
- [ ] New Tool
- [ ] Breaking change
- [x] Other (please describe): Improvements in an existing tool

## Testing
Tested using an example agent here: https://gist.github.com/deshraj/307de1c3385269dda5521edf46567ba0

## Checklist
- [ ] I have read the CONTRIBUTING document
- [ ] My changes generate no new warnings

### DIFF
--- a/src/strands_tools/mem0_memory.py
+++ b/src/strands_tools/mem0_memory.py
@@ -102,8 +102,8 @@ TOOL_SPEC = {
         "- retrieve: Semantic search (requires user_id or agent_id or run_id)\n"
         "- delete: Delete memory\n"
         "- history: Get memory history\n\n"
-        "Note: Most operations require either user_id or agent_id or run_id to be specified. The tool will automatically "
-        "attempt to retrieve relevant memories when user_id or agent_id or run_id is available."
+        "Note: Most operations require either user_id or agent_id or run_id to be specified. The tool will "
+        "automatically attempt to retrieve relevant memories when user_id or agent_id or run_id is available."
     ),
     "inputSchema": {
         "json": {
@@ -136,14 +136,14 @@ TOOL_SPEC = {
                 },
                 "run_id": {
                     "type": "string",
-                    "description": "Run ID / Session ID for the memory operations (required for store, list, retrieve actions)",
+                    "description": "Run/Session ID for memory operations (required for store, list, retrieve actions)",
                 },
                 "metadata": {
                     "type": "object",
                     "description": "Optional metadata to store with the memory",
                 },
             },
-            "required": ["action"]
+            "required": ["action"],
         }
     },
 }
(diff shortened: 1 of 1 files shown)


src/strands_tools/mem0_memory.py
```diff
@@ -216,7 +218,18 @@ def _initialize_opensearch_client(self, config: Optional[Dict] = None) -> Mem0Me
 
         Returns:
             An initialized Mem0Memory instance configured for OpenSearch.
+
+        Raises:
+            ImportError: If opensearchpy package is not installed.
         """
+        try:
+            from opensearchpy import AWSV4SignerAuth, RequestsHttpConnection
```
Unshure, line 226: I had trouble installing `faiss` before, so I could understand that we didnt include it as a top level import, but I dont see anything wrong with including this at the top level. We have it specified in the mem_0 memory optional dependencies here: https://github.com/strands-agents/tools/blob/main/pyproject.toml#L77

src/strands_tools/mem0_memory.py
```diff
@@ -91,20 +90,20 @@
...
-        "2. Retrieve memories by ID or semantic search (requires user_id or agent_id)\n"
-        "3. List all memories for a user/agent (requires user_id or agent_id)\n"
+        "1. Store memories with metadata (requires user_id or agent_id or run_id)\n"
+        "2. Retrieve memories by ID or semantic search (requires user_id or agent_id or run_id)\n"
+        "3. List all memories for a user/agent (requires user_id or agent_id or run_id)\n"
         "4. Delete memories\n"
         "5. Get memory history\n\n"
         "Actions:\n"
-        "- store: Store new memory (requires user_id or agent_id)\n"
+        "- store: Store new memory (requires user_id or agent_id or run_id)\n"
```
Unshure, line 99: Nit: Any reason to write this as `(requires user_id or agent_id or run_id)` instead of `requires user_id, agent_id, or run_id`? Not sure if there is some LLM preference here? The second option seems more readable

src/strands_tools/mem0_memory.py
deshraj: Yeah, I was thinking that we can make mem0 as a default dependency in the tools repo and make faiss and opensearch as optional so that anyone who only wants to use the mem0 platform, then they don't have to install faiss or opensearch. Let me know what you think.

Unshure: If someone wants to use the mem0 tool, they should be able to use this optional dependency, and have the tool just work. I had trouble installing faiss, so im guessing others might as well, which is why im fine leaving that out with a runtime check. But for the other two, we should just include them in the optional dependency on mem0. 

Since we already have this optional dependency for mem0, i'm hesitant to remove it, and I dont think it makes sense to move the mem0 dependency to a default if we already have this optional dependency.

Unshure:
Closing this as comments have not been responded to in some time. Please feel free to re-open.

PR #101, opened 2025-06-26
Add stability tool

satsumas:
## Description
Adds a tool that will create images using the Stability AI Platform

## Documentation PR
n/a. Please let me know if we should provide this.

## Type of Change
- [ ] Bug fix
- [x] New Tool
- [ ] Breaking change

## Testing
All of the following pass:

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

## Checklist

### DIFF
--- a/src/strands_tools/generate_image_stability.py
+++ b/src/strands_tools/generate_image_stability.py
@@ -32,12 +32,8 @@ from strands_tools import generate_image_stability
 os.environ['STABILITY_API_KEY'] = 'sk-xxx'
 os.environ['STABILITY_MODEL_ID'] = 'stability.stable-image-ultra-v1:1'
 
-If you want to save the generated images to disk, set the environment variable `SAVE_IMAGE` to any
-of "true", "yes", "1".
-os.environ['SAVE_IMAGE'] = 'true'
-
-To specify an output directory for saved images, set the environment variable `STABILITY_OUTPUT_DIR`
-to a local directory path.
+If you want to save the generated images to disk, set the environment variable `STABILITY_OUTPUT_DIR`
+to a local directory where the images should be saved.
 
 If no model is selected, the tool defaults to 'stability.stable-image-core-v1:1'.
 
@@ -330,7 +326,7 @@ def generate_image_stability(tool: ToolUse, **kwargs: Any) -> ToolResult:
     Environment Variables:
         STABILITY_API_KEY: Your Stability Platform API key (required)
         STABILITY_MODEL_ID: The model to use (optional, defaults to stability.stable-image-core-v1:1)
-        SAVE_IMAGE: If set to any value (or "true", "yes", "1"), saves generated images to disk
+        STABILITY_OUTPUT_DIR: If set, saves generated images to disk in the specified directory
 
     Args:
         tool: ToolUse object containing the parameters for image generation.
@@ -402,8 +398,8 @@ def generate_image_stability(tool: ToolUse, **kwargs: Any) -> ToolResult:
         filename = None
         save_info = ""
         # Check if we should save the image to a file
-        save_image = os.environ.get("SAVE_IMAGE", "").lower()
-        if save_image in ("true", "yes", "1"):
+        output_dir = os.environ.get("STABILITY_OUTPUT_DIR")
+        if output_dir:
             # Create a unique filename
             import datetime
             import hashlib
@@ -419,7 +415,6 @@ def generate_image_stability(tool: ToolUse, **kwargs: Any) -> ToolResult:
             unique_id = str(uuid.uuid4())[:6]
 
             # Create directory if it doesn't exist
-            output_dir = os.environ.get("STABILITY_OUTPUT_DIR", "stability_images")
             os.makedirs(output_dir, exist_ok=True)
 
             # Construct the filename
@@ -438,9 +433,10 @@ def generate_image_stability(tool: ToolUse, **kwargs: Any) -> ToolResult:
             "source": {"bytes": image_bytes},
         }
 
+        # Disabled until strands-agents/sdk-python#341 is addressed
         # Add filename to the image object if available
-        if filename:
-            image_object["filename"] = filename
+        # if filename:
+        #    image_object["filename"] = filename
 
         return {
             "toolUseId": tool_use_id,
--- a/tests-integ/test_generate_image_stability.py
+++ b/tests-integ/test_generate_image_stability.py
@@ -1,6 +1,8 @@
 import base64
 import os
+import re
 import shutil
+from unittest import mock
 
 import pytest
 from strands import Agent
@@ -14,6 +16,25 @@ Export it as an environment variable:
 export STABILITY_API_KEY="your_api_key"
 """
 
+if "STABILITY_API_KEY" not in os.environ:
+    pytest.skip(allow_module_level=True, reason="STABILITY_API_KEY environment variable missing")
+
+
+@pytest.fixture(autouse=True)
+def os_environment():
+    keys_to_copy = ["STABILITY_API_KEY", "STABILITY_MODEL_ID"]
+    mock_env = {k: os.environ.get(k) for k in keys_to_copy if k in os.environ}
+    with mock.patch.object(os, "environ", mock_env):
+        yield mock_env
+
+
+@pytest.fixture
+def test_output_dir():
+    path = "test_stability_output"
+    yield path
+    if os.path.exists(path):
+        shutil.rmtree(path)
+
 
 def test_stability_image_generation_integration():
     """
@@ -24,9 +45,6 @@ def test_stability_image_generation_integration():
       - Decode and validate the base64 image
       - Save the output image to a file
     """
-    # Skip if no API key is available
-    if not os.environ.get("STABILITY_API_KEY"):
-        pytest.skip("STABILITY_API_KEY not set - skipping integration test")
 
     # Create agent with the tool
     agent = Agent(tools=[generate_image_stability])
@@ -104,9 +122,6 @@ def test_stability_image_to_image_integration():
     """
     Test image-to-image generation functionality.
     """
-    if not os.environ.get("STABILITY_API_KEY"):
-        pytest.skip("STABILITY_API_KEY not set - skipping integration test")
-
     agent = Agent(tools=[generate_image_stability])
 
     # Create a small test image (1x1 white pixel PNG)
@@ -134,9 +149,6 @@ def test_stability_error_handling_integration():
     """
     Test error handling with invalid inputs.
     """
-    if not os.environ.get("STABILITY_API_KEY"):
-        pytest.skip("STABILITY_API_KEY not set - skipping integration test")
-
     agent = Agent(tools=[generate_image_stability])
 
     # Test with empty prompt (should fail)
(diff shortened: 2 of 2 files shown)


src/strands_tools/generate_image_stability.py
```diff
@@ -0,0 +1,414 @@
...
+import os
+from strands import Agent
+from strands_tools import generate_image_stability
+
+
+# Set your API key and model as environment variables
+os.environ['STABILITY_API_KEY'] = 'sk-xxx'
+os.environ['STABILITY_MODEL_ID'] = 'stability.stable-image-ultra-v1:1'
+
+If no model is selected, the tool defaults to 'stability.stable-image-core-v1:1'.
```
zastrowm, line 38: In the other `generate_image` tool, the model-id is taken in as a parameter so that the [caller/LLM can choose which model to use at runtime](https://github.com/strands-agents/tools/blob/main/src/strands_tools/generate_image.py#L91).  In contrast, this makes the choice-up front.

Both are valid choices and it depends on the use case of the tools, but just figured I'd call out the difference in case you wanted to prefer one or the other.

src/strands_tools/generate_image_stability.py
zastrowm, line 1: > Documentation PR
> n/a. Please let me know if we should provide this.

Two places you might want to provide documentation:

 1. The tools overview section of the readme - helps with discoverability: https://github.com/strands-agents/tools/blob/main/README.md#tools-overview
 2. The Available Tools section of the documentation (visible [here](https://strandsagents.com/latest/user-guide/concepts/tools/example-tools-package/#available-tools) and documented [here](https://github.com/strands-agents/docs/blob/main/docs/user-guide/concepts/tools/example-tools-package.md))

(1) will help with discoverability; (2) is technically optional but will also make the tool more visible.

src/strands_tools/generate_image_stability.py
```diff
@@ -0,0 +1,414 @@
...
+1. Image Generation:
+   • Text-to-image and image-to-image conversion
+   • Support for multiple Stability AI models
+   • Customizable generation parameters (seed, cfg_scale, aspect_ratio)
+   • Style preset selection for consistent aesthetics
+   • Flexible output formats (JPEG, PNG, WebP)
+
+2. Response Format:
+   • Rich response with both text and image data
+   • Status tracking and error handling
```
zastrowm: What does:

> Status tracking and error handling

indicate here?

----

edit: I see this may have been taken from the generate_image tool.

tests/test_generate_image_stability.py
zastrowm, line 1: Thanks for the unit tests!

Would you be open to writing an integration test as well? - conditionally skipped if there's no API key in the environment?  That would allow us to test via CI/CD going forward, which is something that we're in the process of setting up.

The only example we have right now is at https://github.com/strands-agents/tools/blob/main/tests-integ/test_memory_tool.py#L49 though it's a bit more complicated because of the setup required.  The SDK has some examples of conditionally ignoring tests: https://github.com/strands-agents/sdk-python/blob/main/tests-integ/test_model_anthropic.py#L45

zastrowm (approved):
No blockers - have some open questions/callouts.

Possible updates:

 - Updating the readme to callout the new tool (highly suggested)
 - Adding an integration test (nice-to-have, but not a blocker)

src/strands_tools/generate_image_stability.py
satsumas: Yes, I wanted to avoid having the LLM make decisions that could have financial consequences, since different models have different costs. If this is a valid option, let's stick with it.

src/strands_tools/generate_image_stability.py
satsumas: Thanks! I made a PR against the docs here: https://github.com/strands-agents/docs/pull/107/files#

Covered (2). I'll update it to add your suggestion (1) as well.

src/strands_tools/generate_image_stability.py
satsumas: Yeah, I can say something more informative here. We actually return finish_reason, which will indicate where requests have been blocked due to content moderation. I updated the comment to indicate this.

tests/test_generate_image_stability.py
satsumas: OK, I added some integration tests -- please take a look.

I got them to write images to file (always overwriting the same file, so images won't accumulate). If that will cause a problem in CI I can remove this functionality. But sometimes it's reassuring to see an actual output!

src/strands_tools/generate_image_stability.py
```diff
@@ -0,0 +1,414 @@
...
+"""
+
+import base64
+import os
+from typing import Any, Tuple, Union
+
+import requests
+from strands.types.tools import ToolResult, ToolUse
+
+TOOL_SPEC = {
```
zastrowm, line 63: I was testing the tool locally - it works well!

One large difference compared to the `generate_image` tool is that this one does not write the image to the file system.  This works if the tool is used in larger system/workflow, but makes it harder (and uses more tokens) if using it as a local agent. 

Specifically using this workflow (mimicking a CLI workflow):

```
agent = Agent(tools=[generate_image_stability])
agent("Generate an image of a cat")
```

I wasn't able to easily get the image out of the agent.  Would you be open to adding file writing into the tool - optionally gated by an environment variable if you don't want to unconditionally do it?

That would enable something like:

```
agent = Agent(tools=[generate_image_stability])
agent("Generate an image of a cat, then tell me where it was saved")
```

tests/test_generate_image_stability.py
zastrowm: Agreed on seeing the output :) 

Once we get CI/CD up and running, we can adjust as needed.  Thanks for adding these!

src/strands_tools/generate_image_stability.py
satsumas: Thanks @zastrowm.

I think having this feature conditional on an env var makes sense. We've also had trouble getting the OpenAI agent to return images when using the tool. I am discussing the intended behaviour with a wider group, so I'll add my changes once we've come to an agreement.

sayanc82: Thanks @zastrowm having a env var to control the side effect of writing out to a local file makes sense to me. 
However, I think in a slightly longer term we need a way to expose the tool result to the strands agent api user. Also maybe some controls in strands to mark whether the tool output should be fed back into the model or just correlated to the context and passed back to the user ? 
I am trying to write this small story teller application where I need the output from both the model and the tool to create a pdf of text and images, and currently there is no way to do this. I think there is a mechanism to define callback handlers but thats not documented very well and it is also hard to get to the text output from the llm or the prompt used for it from the callback handler. 

Will you be willing to consider a feature request on 
1. Allowing the tool results to be exposed to the application using the strands api.
2. Having a mechanism to control whether tool results need to be passed back into the llm input context or passed directly to the application.
3. Allowing a way to correlated llm output and tool output for the application using the strands api ? 
Maybe a way to do this exists, but it is not clear to me, do let me know.

sayanc82: Specifically I am trying to use the agent this way 

```
agent = Agent(
        model=model, 
        tools=[generate_image],
        system_prompt=system_prompt,
    )
```

and the system prompt is something like 
```
system_prompt = (
        f"You are a children's storybook writer for ages {age_group}. "
        f"Write a story in a {tone} tone with 3 pages. "
        "For each page, output: 'text' (1–2 sentences) and an 'image_prompt'. The image_prompt should be a detailed description of the "
        "image to be generated so that it can be used with a stable diffusion model "
        "Use the image_prompt to generate an image using the generate_image tool. "
        "Return as JSON with 'title' and 'pages'. with the text and the image for each page."
    )
```
and then use the agent like so 

```
prompt = "A tiny cloud who wants to become a rainbow"
 age_group = "4-6"
 tone = "hopeful and magical"

# replace the age_group and tone into the system prompt template above
# create the agent and invoke it with the prompt
# extract the image generated by the tool and the text by the llm and create a pdf book using pdf tools.
response = agent(prompt)
```

src/strands_tools/generate_image_stability.py
```diff
@@ -0,0 +1,414 @@
+"""
+Image generation tool for Strands Agent using Stability Platform API.
+
+This module provides functionality to generate high-quality images using Stability AI's
+latest models including SD3.5, Stable Image Ultra, and Stable Image Core through the
+Stability Platform API.
+
+This means agents can create images in a cost effictive way, using state of the art models.
```
zastrowm:
```suggestion
This means agents can create images in a cost effective way, using state of the art models.
```

src/strands_tools/generate_image_stability.py
zastrowm: @sayanc8, feature requests are always welcome, especially with concrete use cases :) 

That said,  All tool results are available in the agent messages array.  So, for example, generating an image, I was able to get the generated image doing this:


```python
agent = Agent(tools=[generate_image_stability])
result = agent("Generate an image of a cat")
# in theory, it's better to look for a message that has a content with a key of `toolResult`, but for example purposes
generated_image_bytes = agent.messages[-2]['content'][0]['toolResult']['content'][1]['image']['source']['bytes']
```

zastrowm: @satsumas

> We've also had trouble getting the OpenAI agent to return images when using the tool. I am discussing the intended behaviour with a wider group, so I'll add my changes once we've come to an agreement.

Let me know if you need any assistance.

src/strands_tools/generate_image_stability.py
```diff
@@ -0,0 +1,414 @@
...
+    prompt: str,
+    model_id: str,
+    stability_api_key: str,
+    return_type: str = "json",
+    aspect_ratio: str = "1:1",
+    cfg_scale: float = 4.0,
+    seed: int = 0,
+    output_format: str = "png",
+    style_preset: str = None,
+    image: str = None,
```
zastrowm: I think a couple of these should be marked as `Optional[str]` given their default values

src/strands_tools/generate_image_stability.py
sayanc82: thanks a lot @zastrowm !! 
I was looking at response object and not the agent (couldn't figure it out from the doc https://strandsagents.com/latest/api-reference/agent/) . 

I just tried your suggestion and I think I can make it work by making my app generate the book one page at a time. I still think it is good if we we can have a correlation between the text output from the llm and the tool output (and the ability to prevent strands from passing the tool output in its entirety to the context of the llm in the next iteration ) in the messages structures, but this is merely a good to have and not must have as far as the app I am trying to develop is concerned.

Now that I understand it better, 
@satsumas maybe as a followup to the suggestion @zastrowm has, we can add a json object to our response with the exact path where the image is saved (so that the app doesnt have to parse it out of the text response ?
Doesn't get this PR. For this PR I think we are good if we do the env var suggestion @zastrowm has.

sayanc82: >We've also had trouble getting the OpenAI agent to return images when using the tool. I am discussing the >intended behaviour with a wider group, so I'll add my changes once we've come to an agreement.

>>Let me know if you need any assistance.


This is more of a bug report. Any tool which produces a binary/non text output will have this issue with strand and openai. See the detailed error message below. It looks like strands need to massage the tool result into a user message before passing it to openai. 

```
Traceback (most recent call last):
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 201, in event_loop_cycle
    return _handle_tool_execution(
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 394, in _handle_tool_execution
    return recurse_event_loop(
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 290, in recurse_event_loop
    ) = event_loop_cycle(**kwargs)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 171, in event_loop_cycle
    raise e
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 138, in event_loop_cycle
    for event in stream_messages(model, system_prompt, messages, tool_config):
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/streaming.py", line 326, in stream_messages
    yield from process_stream(chunks, messages)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/streaming.py", line 281, in process_stream
    for chunk in chunks:
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/types/models/model.py", line 139, in converse
    for event in response:
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/models/openai.py", line 94, in stream
    response = self.client.chat.completions.create(**request)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/_utils/_utils.py", line 287, in wrapper
    return func(*args, **kwargs)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/resources/chat/completions/completions.py", line 925, in create
    return self._post(
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/_base_client.py", line 1249, in post
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/_base_client.py", line 1037, in request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'messages[3]'. Image URLs are only allowed for messages with role 'user', but this message with role 'tool' contains an image URL.", 'type': 'invalid_request_error', 'param': 'messages[3]', 'code': 'invalid_value'}}
```

src/strands_tools/generate_image_stability.py
sayanc82: >We've also had trouble getting the OpenAI agent to return images when using the tool. I am discussing the >intended behaviour with a wider group, so I'll add my changes once we've come to an agreement.

>>Let me know if you need any assistance.


This is more of a bug report. Any tool which produces a binary/non text output will have this issue with strand and openai. See the detailed error message below. It looks like strands need to massage the tool result into a user message before passing it to openai. 

```
Traceback (most recent call last):
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 201, in event_loop_cycle
    return _handle_tool_execution(
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 394, in _handle_tool_execution
    return recurse_event_loop(
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 290, in recurse_event_loop
    ) = event_loop_cycle(**kwargs)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 171, in event_loop_cycle
    raise e
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/event_loop.py", line 138, in event_loop_cycle
    for event in stream_messages(model, system_prompt, messages, tool_config):
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/streaming.py", line 326, in stream_messages
    yield from process_stream(chunks, messages)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/event_loop/streaming.py", line 281, in process_stream
    for chunk in chunks:
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/types/models/model.py", line 139, in converse
    for event in response:
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/strands/models/openai.py", line 94, in stream
    response = self.client.chat.completions.create(**request)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/_utils/_utils.py", line 287, in wrapper
    return func(*args, **kwargs)
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/resources/chat/completions/completions.py", line 925, in create
    return self._post(
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/_base_client.py", line 1249, in post
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
  File "/Users/sayan.chowdhury/.pyenv/versions/3.10.17/lib/python3.10/site-packages/openai/_base_client.py", line 1037, in request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'messages[3]'. Image URLs are only allowed for messages with role 'user', but this message with role 'tool' contains an image URL.", 'type': 'invalid_request_error', 'param': 'messages[3]', 'code': 'invalid_value'}}
```

sayanc82: > @sayanc8, feature requests are always welcome, especially with concrete use cases :)
> 
> That said, All tool results are available in the agent messages array. So, for example, generating an image, I was able to get the generated image doing this:
> 
> ```python
> agent = Agent(tools=[generate_image_stability])
> result = agent("Generate an image of a cat")
> # in theory, it's better to look for a message that has a content with a key of `toolResult`, but for example purposes
> generated_image_bytes = agent.messages[-2]['content'][0]['toolResult']['content'][1]['image']['source']['bytes']
> ```

thanks a lot @zastrowm !!
I was looking at response object and not the agent (couldn't figure it out from the doc https://strandsagents.com/latest/api-reference/agent/) .

I just tried your suggestion and I think I can make it work by making my app generate the book one page at a time. I still think it is good if we we can have a correlation between the text output from the llm and the tool output (and the ability to prevent strands from passing the tool output in its entirety to the context of the llm in the next iteration ) in the messages structures, but this is merely a good to have and not must have as far as the app I am trying to develop is concerned.

Now that I understand it better,
@satsumas maybe as a followup to the suggestion @zastrowm has, we can add a json object to our response with the exact path where the image is saved (so that the app doesnt have to parse it out of the text response ?
Doesn't gate this PR. For this PR I think we are good if we do the env var suggestion @zastrowm has

zastrowm: > This is more of a bug report. Any tool which produces a binary/non text output will have this issue with strand and openai. See the detailed error message below. It looks like strands need to massage the tool result into a user message before passing it to openai.

Indeed, it seems like a limitation of the OpenAI tool functions returning images: 

 - [Returning image as tool output in Assistants API? - API - OpenAI Developer Community](https://community.openai.com/t/returning-image-as-tool-output-in-assistants-api/763387/5)
 - [Returning image as result of function call to gpt-4-turbo - API / Bugs - OpenAI Developer Community](https://community.openai.com/t/returning-image-as-result-of-function-call-to-gpt-4-turbo/714903/10)

So we probably need to update the OpenAI provider to workaround this somehow - I'll cut a bug for that.

satsumas: @zastrowm, please see the latest changes for the image save implementation.

BTW, will you run integration tests that let agents based on each model use the tool? For instance, I am finding that intermittently the Anthropic-model-based agent will be unable to return the image because it exceeds an upper limit of 5MB for responses (error below). 

Stability's images are 1MP for SD3.5 and Ultra, and 1.5MP for core. Base64-encoding increases filesize by 33%. It's not obvious to me how we're getting from 1.5MP images to >5MB response bodies -- I will ask internally. Do you have any recommendations on how to proceed so that the Stability tool is compatible with all models, or else to document which ones it should be used with"?

```
raise EventLoopException(e, kwargs["request_state"]) from e
strands.types.exceptions.EventLoopException: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.2.content.0.tool_result.content.1.image.source.base64: image exceeds 5 MB maximum: 5586388 bytes > 5242880 bytes'}}
```

I can move this latter discussion to an issue.

zastrowm: Yes, I will run integ tests and report back;  I'll also take a look to see if there's anything on our hand that might be bloating the response

zastrowm: > This is more of a bug report. Any tool which produces a binary/non text output will have this issue with strand and openai. See the detailed error message below. It looks like strands need to massage the tool result into a user message before passing it to openai.

FYI, filed this as [[BUG] OpenAI provider cannot use tools that return images · Issue #320 · strands-agents/sdk-python](https://github.com/strands-agents/sdk-python/issues/320)

satsumas: I resolved an issue in my local dev environment and since then have not seen the `image exceeds 5 MB maximum` error.

So we can probably safely proceed on the assumption that this is not an issue. But should ensure that any model provider that will be used in conjunction with the Stability tool is able to return at least 2MB-sized responses. I will add this to the tool documentation.

zastrowm (approved):

PR #49, opened 2025-05-26
feat: Add MCP client tool

mkmeral:
## Description
This PR introduces a new MCP (Model Context Protocol) client tool for Strands Agents that provides a high-level interface for connecting to any MCP server with simplified configuration and enhanced functionality.

### Key Features:
- **Multiple Transport Support**: Connect to MCP servers via stdio (for local processes) or SSE (Server-Sent Events) for web-based servers
- **Simplified Configuration**: Direct parameter passing without complex nested configurations
- **Connection Management**: Thread-safe connection storage with support for multiple simultaneous connections
- **Tool Discovery**: List available tools from connected MCP servers
- **Direct Tool Invocation**: Call MCP tools directly or load them into the agent's tool registry
- **Environment Variable Support**: Pass environment variables to stdio-based MCP servers
- **Comprehensive Error Handling**: Detailed error messages and connection status tracking

### Usage Examples:
```python
# Connect to a custom MCP server
agent.tool.mcp_client(
    action="connect",
    connection_id="my_server",
    transport="stdio",
    command="python",
    args=["server.py"]
)

# List available tools
agent.tool.mcp_client(
    action="list_tools",
    connection_id="my_server"
)

# Call a tool directly
agent.tool.mcp_client(
    action="call_tool",
    connection_id="my_server",
    tool_name="calculate",
    x=10,
    y=20
)

# Load tools into agent's registry for direct access
agent.tool.mcp_client(
    action="load_tools",
    connection_id="my_server"
)
```

## Related Issues
N/A

## Documentation PR
N/A - Documentation is included in the README.md update

## Type of Change
- [ ] Bug fix
- [x] New Tool
- [ ] Breaking change
- [ ] Other (please describe):

## Testing
The implementation includes comprehensive unit tests covering:
- Connection management (stdio and SSE transports)
- Tool listing and discovery
- Direct tool invocation
- Tool loading into agent registry
- Error handling scenarios
- Thread safety
- Full workflow integration tests

Testing commands run:
* `hatch fmt --linter` ✓
* `hatch fmt --formatter` ✓
* `hatch test --all` ✓

#### Manual Testing
```
**user:** use perplexity mcp from ~/.aws/amazonq/mcp.json and tell me what techno events are in amsterdam this weekend

**strands:** connected to Perplexity MCP server using `mcp_client(action="connect")` with stdio transport, reading config from ~/.aws/amazonq/mcp.json

[...]

PR #108, opened 2025-07-01
feat(a2a): native client tool

jer96:
## Description
This PR creates a native A2A client tool. This tool allows a strands agent to discover and invoke A2A enabled servers. The tool also enables developers to provide a list of known agent urls to initialize the tool directly with a list of A2A enabled servers.

## Related Issues
https://github.com/strands-agents/private-sdk-python-staging/issues/96

## Documentation PR
Upcoming: https://github.com/strands-agents/private-sdk-python-staging/issues/98

## Type of Change
- [ ] Bug fix
- [x] New Tool
- [ ] Breaking change

## Testing
- Unit tests
- Manually tested with sample agent and A2A server running locally

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

Sample scripts
a2a_server.py
```python 
import logging
import sys

from strands import Agent
from strands.multiagent.a2a import A2AServer

from strands_tools.calculator import calculator

logging.basicConfig(
    level=logging.INFO,
    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
    handlers=[logging.StreamHandler(sys.stdout)],
    force=True,
)

# Log that we're starting
logging.info("Starting A2A server with root logger")

strands_agent = Agent(
    model="us.anthropic.claude-3-haiku-20240307-v1:0",
    name="Calculator Agent",
    tools = [calculator],
    description="A calculator agent that can perform basic arithmetic operations.",
    callback_handler=None
)
strands_a2a_agent = A2AServer(agent=strands_agent)
strands_a2a_agent.serve()
```

a2a_client.py
```python
import asyncio
import logging
import sys

from strands import Agent

from strands_tools.a2a_client import A2AClientToolProvider

logging.basicConfig(
    level=logging.INFO,
    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
    handlers=[logging.StreamHandler(sys.stdout)],
    force=True,
)

# sync example 
provider = A2AClientToolProvider(known_agent_urls=["http://localhost:9000"])
agent = Agent(tools=provider.tools)
print(agent("pick an agent and make a sample call"))

# async example
async def main():
    provider = A2AClientToolProvider(known_agent_urls=["http://localhost:9000"])
    agent = Agent(tools=provider.tools)
    await agent.invoke_async("pick an agent and make a sample call")
asyncio.run(main())
```

a2a_client.py output
```
## Summary of the Calculator Agent

I've successfully tested the Calculator Agent which is available at http://0.0.0.0:9000/. This agent can perform various mathematical operations:

1. First test: I performed a basic arithmetic calculation `(5 + 3) * 4 - 2 = 30`
2. Second test: I found the derivative of a polynomial expression `x^3 + 2x^2 - 5x + 1`, which resulted in `3*x^2 + 4*x - 5`

The agent is capable of much more according to its description:
- Different operation modes including evaluate, solve equations, calculate derivatives, find integrals, calculate limits, generate series expansions, and perform matrix operations
- Support for variables and constants
- Precision control and scientific notation
- Symbolic and numeric evaluations

If you'd like to try something more complex with this agent or explore other aspects of its functionality, please let me know!
2025-07-03 13:14:21,444 - strands_tools.a2a_client - INFO - A2AClientToolProvider resources cleaned up
```

## Checklist


src/strands_tools/base/tool_provider.py
```diff
@@ -0,0 +1,19 @@
...
+class ToolProvider(ABC):  # noqa: B024
+    """Base class for creating stateful tool providers."""
+
+    def get_tools(self) -> list[AgentTool]:
+        """Extract all @tool decorated methods from this instance."""
+        tools = []
+
+        for attr_name in dir(self):
+            attr = getattr(self, attr_name)
+            if isinstance(attr, DecoratedFunctionTool):
```
zastrowm: I think we should be checking for `AgentTool` instead.

src/strands_tools/base/tool_provider.py
```diff
@@ -0,0 +1,19 @@
+from abc import ABC
+
+from strands.tools.decorator import DecoratedFunctionTool
+from strands.types.tools import AgentTool
+
+
+class ToolProvider(ABC):  # noqa: B024
+    """Base class for creating stateful tool providers."""
+
+    def get_tools(self) -> list[AgentTool]:
```
zastrowm: For now, can we keep this isolated to the a2a tool - this feels like something that should live in the SDK if we want it, but I'd like to see enough usage of this to warrant it. Esp because we're being opinionated on how get_tools is implemented

src/strands_tools/base/tool_provider.py
```diff
@@ -0,0 +1,19 @@
+from abc import ABC
+
+from strands.tools.decorator import DecoratedFunctionTool
```
zastrowm: This actually probably indicates that `DecoratedFunctionTool` needs to be exposed from `strands.tools`, but honestly I'd like to keep it as an implementation detail

src/strands_tools/a2a_client.py
```diff
@@ -0,0 +1,248 @@
...
+
+        resolver = await self._create_a2a_card_resolver(url)
+        agent_card = await resolver.get_agent_card()
+        self._discovered_agents[url] = agent_card
+        logger.info(f"Successfully discovered and cached agent card for {url}")
+
+        return agent_card
+
+    @tool
+    def discover_agent(self, url: str) -> dict[str, Any]:
```
zastrowm, line 136: Check with @pgrayy, but I wonder if we should be leaning into async tools

src/strands_tools/a2a_client.py
```diff
@@ -0,0 +1,248 @@
...
+            discover_on_init: Whether to discover all input agent cards during initialization (defaults to False)
+        """
+        self.timeout = timeout
+        self._agent_urls: list[str] = agent_urls or []
+        self._discovered_agents: dict[str, AgentCard] = {}
+        self._httpx_client: httpx.AsyncClient | None = None
+        self._cleanup_registered = False
+
+        # Discover agents on initialization if requested
+        if discover_on_init and self._agent_urls:
```
zastrowm: It seems like if `discover_on_init` is specified but `_agent_urls` is not, it's most likely user error and I think we should throw?

src/strands_tools/a2a_client.py
```diff
@@ -0,0 +1,248 @@
...
+        atexit.register(self.close)
+
+    def _run_async(self, coro):
+        """Handle async-to-sync conversion internally."""
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+        return loop.run_until_complete(coro)
```
zastrowm, line 81: Discuss with @pgrayy - I thought we found that if a thread is in an event loop/context, then this won't work?

src/strands_tools/a2a_client.py
```diff
@@ -0,0 +1,248 @@
...
+        except Exception as e:
+            logger.exception("Error listing discovered agents")
+            return {
+                "status": "error",
+                "error": str(e),
+                "total_count": 0,
+            }
+
+    @tool
+    def send_message(self, message_text: str, target_agent_url: str, message_id: str | None = None) -> dict[str, Any]:
```
zastrowm, line 208: I think the proposed mcp tool (#49) and the use_browser tool (#88) all use overloaded tools with actions instead of individual messages. 

Did we want to mimic that and/or what are the pros and cons?

src/strands_tools/a2a_client.py
```diff
@@ -0,0 +1,248 @@
...
+        except Exception as e:
+            logger.exception(f"Error sending message to {target_agent_url}")
+            return {
+                "status": "error",
+                "error": str(e),
+                "message_id": message_id,
+                "target_agent_url": target_agent_url,
+            }
+
+    def close(self):
```
Unshure, line 267: Can we call close in `__exit__`: https://docs.python.org/2/reference/datamodel.html#with-statement-context-managers

src/strands_tools/a2a_client.py
jer96: yep agreed

src/strands_tools/base/tool_provider.py
jer96: yep will remove

src/strands_tools/a2a_client.py
jer96: i've implemented this in a simliar way. it's actually implemented using `atexit` [above](https://github.com/strands-agents/tools/pull/108/files#diff-cf28e6b27cb77ae480c5b7d520b455412baf154dfb3494a5c1266bdb187e0353R58)

src/strands_tools/a2a_client.py
jer96: i would love to lean into async tools, in many ways that would simplify this PR greatly and i could remove a solid amount of code. @pgrayy any recommendation here?

src/strands_tools/a2a_client.py
```diff
@@ -0,0 +1,248 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class A2AClientToolProvider(ToolProvider):
+    """A2A Client tool provider that manages multiple A2A agents and exposes synchronous tools."""
+
+    def __init__(
+        self,
+        agent_urls: list[str] | None = None,
```
pgrayy: Nit: Conventionally, an optional type is denotated as `: Optional[list[str]] = None`

tests/test_a2a_client.py
```diff
@@ -0,0 +1,342 @@
...
+"""
+
+from unittest.mock import AsyncMock, Mock, patch
+
+import pytest
+from a2a.types import Role, SendMessageRequest
+from strands_tools.a2a_client import DEFAULT_TIMEOUT, A2AClientToolProvider
+
+
+class TestA2AClientToolProvider:
```
pgrayy: Nit: pytest doesn't require the class namespace like unittest, so you can save on some indentation by removing it.

src/strands_tools/a2a_client.py
pgrayy: Consider the following example:
```Python
import asyncio


async def func_a():
    return 1


def func_b():
    loop = asyncio.get_event_loop()
    return loop.run_until_complete(func_a())


async def func_c():
    return func_b()


asyncio.run(func_c())
```

This fails with:
```txt
RuntimeError: This event loop is already running
```

I don't think you can get around this without spinning up a new thread (for an example, see https://github.com/strands-agents/sdk-python/blob/eb1ffd542d0c1fa222dc384d8a6b980efbaedc10/src/strands/agent/agent.py#L374). There may be a workaround I haven't discovered though.

Whatever the approach, I would recommend trying to capture something like this in your unit tests using `@pytest.mark.asyncio` to create the async context.

src/strands_tools/a2a_client.py
zastrowm: I've commented this a couple times on the PRs enough that I also did research on it: [Python 3.10+: Optional[Type] or Type | None - Stack Overflow](https://stackoverflow.com/questions/69440494/python-3-10-optionaltype-or-type-none)

----

Honestly I prefer the ` | ` syntax and think we should switch the whole code base over at some point, though for now I prefer consistency with `Union` and `Optional` unless someone will do the conversation (nerd snipe!)

src/strands_tools/a2a_client.py
zastrowm: FWIW, I think at some point we should have a strands helper to aid in this.  Something like:

```python
run_async_code_sync(callable)
```

That way if we have some optimizations later, we benefit in all the places

jer96: @pgrayy i tested your recommended approach with this PR and it actually causes errors. specifically: 

```
  raise RuntimeError(f'{self!r} is bound to a different event loop')
RuntimeError: <asyncio.locks.Event object at 0x107c146b0 [unset]> is bound to a different event loop
```

i have opted for the initial strategy of obtaining the event loop and it does indeed work. moreover, i've tested with your latest changes to `__call__` and `invoke_async` and the tool functions properly in both settings.


```python
import asyncio
import logging
import sys

from strands import Agent

from strands_tools.a2a_client import A2AClientToolProvider

provider = A2AClientToolProvider(known_agent_urls=["http://localhost:9000"])
agent = Agent(tools=provider.tools)
print(agent("pick an agent and make a sample call"))

async def main():
    provider = A2AClientToolProvider(known_agent_urls=["http://localhost:9000"])
    agent = Agent(tools=provider.tools)
    await agent.invoke_async("pick an agent and make a sample call")
asyncio.run(main())

```

src/strands_tools/a2a_client.py
jer96: i spoke offline with @pgrayy and i've opted to move away from performing any async operations in the constructor and move discovery of the known agent urls into the actual tool invocations

src/strands_tools/a2a_client.py
jer96: since this is separate from the sdk-python package and a new tool can we prefer the new syntax? iirc all of our repos min version is 3.10, so imo using the pipe syntax would be preferred.

src/strands_tools/a2a_client.py
jer96: in truth, i've only browsed the other tools briefly, but from a quick glance, i feel like this does add sufficient complexity and reduces the readability of the tool. i'd also venture to say that in some ways it makes the code less maintainable since arbitrary objects can be passed in as actions. 


i personally prefer self contained tools with the `@tool` decoration as in this PR. i do think this warrants more investigation, but i would prefer to leave that out of scope of this PR.

pyproject.toml
```diff
@@ -81,86 +79,64 @@ mem0_memory = [
     "mem0ai>=0.1.99,<1.0.0",
     "opensearch-py>=2.8.0,<3.0.0",
 ]
+a2a = ["a2a-sdk>=0.2.6"]
```
zastrowm, line 82: I think we've started naming the feature to match the tool name; could we do that here?

```
a2a_client = 
```

(LMK if this doesn't make sense though)

src/strands_tools/a2a_client.py
pgrayy: I think this is because the tool is running in a separate thread. Can you try setting max_parallel_tools=1 in the agent init?

jer96: i set `max_parallel_tools=1` and i did encounter the error. based on our offline conversation, i've updated the documentation to communicate avoiding `max_parallel_tools=1` and i will fast follow a change to async tools once it's been implemented.

PR #113, opened 2025-07-02
test(http_request): add integration test for http_request tool

JackYPCOnline:
## Description
[Provide a detailed description of the changes in this PR]
add serval tests for http_request tool.

## Related Issues

## Documentation PR

## Type of Change
- [ ] Bug fix
- [ ] New Tool
- [ ] Breaking change
- [ x] Other (please describe): Test

## Testing

* `hatch fmt --linter`
* `hatch fmt --formatter`
* `hatch test --all`

## Checklist
- [x ] I have read the CONTRIBUTING document
- [x ] I have added tests that prove my fix is effective or my feature works
- [x ] I have updated the documentation accordingly
- [x ] I have added an appropriate example to the documentation to outline the feature
- [x ] My changes generate no new warnings
- [x ] Any dependent changes have been merged and published

### DIFF
--- a/tests-integ/test_http_request.py
+++ b/tests-integ/test_http_request.py
@@ -5,11 +5,18 @@ to handle various HTTP methods (GET, POST, PUT) and operations like downloading
 files and interacting with AWS S3 services.
 """
 
+import json
 import os
+import socket
+import threading
+import time
+from http.server import BaseHTTPRequestHandler, HTTPServer
 
 import pytest
 from strands import Agent
 from strands_tools import http_request
+
+os.environ["BYPASS_TOOL_CONSENT"] = "true"
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


tests-integ/test_http_request.py
```diff
@@ -0,0 +1,92 @@
...
+    Verifies that the agent can successfully make a POST request with a JSON body
+    to httpbin.org and receive the expected response containing the sent data.
+
+    Note:
+        Test may be skipped if the external endpoint is unavailable or returns
+        a cancelled response due to service instability.
+    """
+    response = agent('Send a POST request to https://httpbin.org/post with JSON body {"foo": "bar"}')
+    if "cancelled" in str(response).lower():
+        pytest.skip("This test is flaky due to their endpoint. Skipping for now.")
```
zastrowm: What kind of instability do you see? Is there a more reliable endpoint we could use?

My concern is if this starts failing, would we ever know or would it just be an expected skip case?

zastrowm: Better yet - can we just spin up a python based http echo server that we control for the test?  That way we aren't using public resources and can have more deterministic responses

zastrowm (changes requested):

tests-integ/test_http_request.py
JackYPCOnline: rewrited the code to use self host server to handle the tests.

tests/test_mem0.py
```diff
@@ -336,13 +337,19 @@ def test_invalid_action(mock_opensearch, mock_mem0_client, mock_tool):
...
+
+    real_import = builtins.__import__
+
+    def fail_faiss(name, *args, **kwargs):
+        if name == "faiss":
+            raise ImportError("No module named 'faiss'")
+        return real_import(name, *args, **kwargs)
 
-    # Mock faiss import
-    with patch("strands_tools.mem0_memory.faiss", create=True):
```
JackYPCOnline, line 345: This test fails when user has faiss installed, so rewrite it to unblock PR.

zastrowm (approved):

PR #163, opened 2025-07-22
 Fix Bedrock Knowledge Base retrieval by adding overrideSearchType parameter

pavelsjo:
This PR fixes an issue with the Strands Agent's retrieve tool when accessing Amazon Bedrock Knowledge Bases. The problem was causing "ResourceNotFoundException" errors even when the Knowledge Base existed because the API call was missing a required parameter in the retrievalConfiguration.

The fix ensures compatibility with the current Bedrock Agent Runtime API specification by adding the required "overrideSearchType" parameter to the vectorSearchConfiguration object.

### Changes Made

1. Updated the retrieve.py implementation to include the required "overrideSearchType" parameter in the vectorSearchConfiguration object when calling the Bedrock Agent Runtime API
2. Added explicit AWS region configuration to ensure the tool connects to the correct region where the Knowledge Base is deployed (os.environ["AWS_REGION"] = "us-east-1")
3. Created test scripts to verify proper functioning of the retrieve tool with the updated parameters
4. Added documentation explaining the parameter requirements for the Bedrock Agent Runtime API

### Testing Performed

- Verified Knowledge Base existence using AWS CLI (aws bedrock-agent list-knowledge-bases)
- Successfully retrieved content from the Knowledge Base using the updated tool
- Confirmed proper scoring and filtering of results
- Tested with both direct API calls and through the Strands Agent interface

### -Root Cause Analysis
The issue was caused by changes in the Bedrock Agent Runtime API that now requires the "overrideSearchType" parameter in the vectorSearchConfiguration object. Without this parameter, the API returns a ResourceNotFoundException error even when the Knowledge Base exists.

### Related Issues
"ResourceNotFoundException when using retrieve tool with Bedrock Knowledge Base"

### Type of Change
[x] Bug fix
[ ] New Tool
[ ] Breaking change
[ ] Other (please describe):

### Screenshots/Logs

Before fix:

`Error during retrieval: An error occurred (ResourceNotFoundException) when calling the Retrieve operation: Knowledge Base with id XXXXXXXXX does not exist`

After fix:

```
Retrieved 1 results with score >= 0.4:
Score: 0.4336
Content: Restaurant Directory...
Additional Notes
This change maintains backward compatibility while ensuring the tool works correctly with the current API version. Users should ensure they have the correct AWS region configured for their Knowledge Base.
```

### DIFF
--- a/src/strands_tools/retrieve.py
+++ b/src/strands_tools/retrieve.py
@@ -278,7 +278,10 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
             retrievalQuery={"text": query},
             knowledgeBaseId=kb_id,
             retrievalConfiguration={
-                "vectorSearchConfiguration": {"numberOfResults": number_of_results},
+                "vectorSearchConfiguration": {
+                    "numberOfResults": number_of_results,
+                    "overrideSearchType": "HYBRID"
+                },
             },
         )
 


JackYPCOnline:
Thank you for your contribution! We truly appreciate community members like you taking the time to help improve the project. Based on bedrock [docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_KnowledgeBaseVectorSearchConfiguration.html), overrideSearchType is not a required parameter.

rromanchuk:
https://github.com/strands-agents/tools/blob/874dfdf1a1a3b23bc00fcb87afa13838a654b470/src/strands_tools/retrieve.py#L330C5-L336C10

Ughhhh this is a nightmare to debug, i wish it it wouldn't catch this and just raise. It just makes the llm gaslight you and you have to chase down inter-agent-tool message passing in logs and i'm guessing str(e) gets coerced in the absolute least helpful key

```
{
    'message': 'string',
    'Error': {
        'Code': 'string',
        'Message': 'string'
    }
} 
```

PR #184, opened 2025-07-28
Fix: workflow dependency task returns result content in result.message

vria:
## Description

Using dependencies in the workflow tool was unsuccessful because the results from dependency task were not available in the context of its dependent task. The issue was that the workflow tool did not properly extract the content from the AgentResult object returned by the dependency task.
This PR fixes it.

## Related Issues

None

## Documentation PR

None

## Type of Change

Bug fix
Backward compatible

## Testing

## Checklist
- [ ] I have added any necessary tests that prove my fix is effective or my feature works

----


cagataycali (approved):
thank you for contributing!

src/strands_tools/workflow.py
```diff
@@ -444,12 +444,21 @@ def execute_task(self, task: Dict, workflow: Dict) -> Dict:
...
-                content = (
-                    result.message.get("content", [])
-                    if hasattr(result.message, "get")
-                    else getattr(result.message, "content", [])
-                )
+                content = []
+                if hasattr(result, "get"):
+                    content = result.get("content", [])
+                elif hasattr(result, "content"):
+                    content = getattr(result, "content", [])
```
JackYPCOnline, line 453: getattr already has a default, so hasattr is unnecessary.

Issue #187, opened 2025-07-30
[FEATURE] AgentCore custom browser tools identifier support


stevensu1977:
### Problem Statement

I found AgentCoreBrowser only support default  identifier , I  add custom browser tools identifier  support  .

```python
# I add start method identifier argument.
# Create new browser client for this session
session_client = AgentCoreBrowserClient(region=self.region)
session_id = session_client.start(identifier=self.identifier, session_timeout_seconds=self.session_timeout)


````

### Use Case

AgentCore Custom Browser Tools

PR #188, opened 2025-07-30
Add identifier to AgentCoreBrowser

stevensu1977:
## Description

Add AgentCore custom browser tool identifier support.

## Related Issues

#187 

## Checklist
- [] I have added any necessary tests that prove my fix is effective or my feature works

----


dbschmigelski (approved):
Thanks for this. Adding some color that the implication of this is that when an AgentCoreBrowser is being used it previously would have defaulted to reusing the same instance which is pregenerated for all AWS accounts.

PR #243, opened 2025-09-09
fix: Fix deadlock due to playwright used in a nested event loop

osdemah:
## Description

Playwright instance is not safe to be used across different event loops. The bad practice of using nested event loops causes deadlock if the playwright instance is used for further operations after the agent is done. (e.g. processing cookies, checking history, etc).

Example code to reproduce the deadlock:

```
async def test():
    browser = LocalChromiumBrowser()
    browser._start()
    browser.init_session(InitSessionAction(type="init_session", session_name="session-1234", description="my-frist-session"))
    print(browser.navigate(NavigateAction(type="navigate", session_name="session-1234", url="https://amazon.com")))
    print(browser.get_cookies(GetCookiesAction(type="get_cookies", session_name="session-1234")))
    page = browser.get_session_page("session-1234")
    assert page
    # following line never finishes!
    print(await page.context.cookies())
```

Considering strands support of asyncio, we can now simply call Playwright APIs in the same event loop as agent's. The only exception is the `__del__` function which is itself a bad practice but kept it to avoid resource leakage in the consumers already using this tool. Ideally the consumers should switch to using provided async context manager that calls `__aexit__` when block is finished.

## Type of Change

Bug fix

## Testing

- [X] I ran `hatch run prepare`

## Checklist
- [X] I have added any necessary tests that prove my fix is effective or my feature works

----


src/strands_tools/browser/browser.py
```diff
@@ -132,68 +132,73 @@ def browser(self, browser_input: BrowserInput) -> Dict[str, Any]:
...
+    async def __aenter__(self):
+        await self._start()
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        try:
+            await self._cleanup()
+        except Exception as e:
+            logger.debug("exception=<%s> | platform cleanup during destruction skipped", str(e))
+        await self._cleanup()
```
mkmeral, line 188: there is double cleanup right now, do we need both?

src/strands_tools/browser/browser.py
```diff
@@ -132,68 +132,73 @@ def browser(self, browser_input: BrowserInput) -> Dict[str, Any]:
...
+            return await self.back(action)
         elif isinstance(action, ForwardAction):
-            return self.forward(action)
+            return await self.forward(action)
         elif isinstance(action, RefreshAction):
-            return self.refresh(action)
+            return await self.refresh(action)
         elif isinstance(action, EvaluateAction):
-            return self.evaluate(action)
+            return await self.evaluate(action)
```
mkmeral, line 165: evaluate is a sync method, we shouldn't use await

PR #254, opened 2025-09-18
Add mem0_memory Support for PostgreSQL

ictmalili:
## Description

This PR aims to add support for PostgreSQL as vector store of mem0_memory. Given that mem0 itself supports PostgreSQL as underlying storage, we can enrich the mem0_support for PostgreSQL.

## Related Issues
https://github.com/strands-agents/tools/issues/258

## Type of Change

Other (please describe): New function

## Testing

## Checklist

----


src/strands_tools/mem0_memory.py
```diff
@@ -266,6 +293,37 @@ def _append_neptune_database_backend(self, config: Optional[Dict] = None) -> Dic
 
         return config
 
+    def _initialize_postgresql_client(self, config: Optional[Dict] = None) -> Mem0Memory:
```
javierlarota, line 296: I would use the same naming and code convention like the other `_append*` methods. I would call this method `_append_pgvector_config` 

Also, I would keep setting the `config["vector_store"]` here in this method rather than in a separated method. Again, to follow the same code convention similar to the other `_append` methods

Issue #280, opened 2025-10-09
[BUG] Metadata not coming in response through retrieve tool


SETHJI01:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.11.0

### Tools Package Version

0.1.dev118+g985a75add.d20251009

### Tools used

1. retrieve

### Python Version

3.11.0

### Operating System

macOS 15.7.1

### Installation Method

pip

### Steps to Reproduce

1. Install strands-agent and strands-agent-tools using pip.
2. try to use the retrieve method by creating a agent.
3. check the response from the retrieve method.

### Expected Behavior

metadata is not present
`{"role": "user", "content": [{"toolResult": {"toolUseId": "tooluse_V8unGSoKShqASzdVSybqWA", "status": "success", "content": [{"text": "Retrieved 1 results with score >= 0.6:\n\nScore: 0.6648\nDocument ID: s3://smartcommerce-documents-215362879456-alpha/smartbiz/PaymentProcessing/how-to-manage-gst-and-invoicing.md\nContent: --- id: 35479951624977 title: \"How to manage GST and Invoicing\" product: smartbiz domain: payments tags: [gst, invoicing, tax, billing, compliance] --- # How to manage GST and Invoicing ## Overview SmartBiz's GST Invoice feature enables automatic generation of invoices for your online orders. As a seller, you can set up GST details in your tax settings by providing your GSTIN, business name, and applicable tax rates. Once enabled, invoices will be automatically generated for each order and can be downloaded by both you and your customers. ## Prerequisites - Valid GST registration number (GSTIN) - Business registered name matching GST registration - Business registered address - Understanding of applicable tax rates for your products ## Step-by-Step Instructions ### GST Tax Setting Page You can access this page by clicking on \"Setting\" on left hand side and then selecting \"Tax\". **Required Information:** **a) GST No:** Input your GST number. **b) State Code and State Name:** This information is captured through first two digits of GST number. **c) Business Registered Name:** Input your business registered name of the same GST that was populated in the field earlier. It will be used to populate \"Sold By\" information in the invoice. **d) Business Registered Address:** Input your business registered address of the same GST that was populated in the field earlier. It will be used to populate \"Sold By\" information in the invoice.\n"}]}}]},`

### Actual Behavior

metadata should be present in it

[...]

PR #283, opened 2025-10-10
feat: enhance retrieve tool with configurable metadata output (#280)

DemotivatedEngineer:
## Description
The retrieve tool currently does not support including metadata in its response.
This change introduces an optional parameter that enables metadata ingestion in the retrieve method’s response.

#### Implementation Details:
    - Added enableMetadata boolean parameter to retrieve tool TOOL_SPEC
    - Implemented environment variable support for default metadata behavior
    - Updated result formatting to conditionally display metadata fields
    - Ensured backward compatibility (metadata disabled by default)
    - Added comprehensive unit test coverage
    - Updated documentation and usage examples

## Related Issues

https://github.com/strands-agents/tools/issues/280

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands_tools/retrieve.py
+++ b/src/strands_tools/retrieve.py
@@ -34,6 +34,12 @@ agent = Agent(tools=[retrieve])
 # Basic search with default knowledge base and region
 results = agent.tool.retrieve(text="What is the STRANDS SDK?")
 
+# Search with metadata enabled for source information
+results = agent.tool.retrieve(
+    text="What is the STRANDS SDK?",
+    enableMetadata=True
+)
+
 # Advanced search with custom parameters
 results = agent.tool.retrieve(
     text="deployment steps for production",
@@ -41,6 +47,7 @@ results = agent.tool.retrieve(
     score=0.7,
     knowledgeBaseId="custom-kb-id",
     region="us-east-1",
+    enableMetadata=True,
     retrieveFilter={
         "andAll": [
             {"equals": {"key": "category", "value": "security"}},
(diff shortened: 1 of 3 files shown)


src/strands_tools/retrieve.py
```diff
@@ -34,13 +34,20 @@
 # Basic search with default knowledge base and region
 results = agent.tool.retrieve(text="What is the STRANDS SDK?")
 
+# Search with metadata enabled for source information
+results = agent.tool.retrieve(
```
JackYPCOnline, line 38: Can you explain why we need this to be configurable? https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_Retrieve.html

Also what is the usecase for including this metadata.

DemotivatedEngineer: Hi @JackYPCOnline for reviewing the PR
For my use case, I’m creating a agent to use a knowledge base created using **S3 data source**.

To enable **citations** in model responses, we need the actual URL of the document stored in S3. To achieve this, I use **metadata** to override the `x-amz-bedrock-kb-source-uri`.
(Reference: [AWS Bedrock Documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/kb-metadata.html)

Additionally, the metadata can include other custom parameters, such as `documentType: Public/Private`, which can help control document visibility and restrict access for unauthorized users.

I’ve made this metadata feature **configurable**, since not every use case requires it.

Lastly, as noted in [this line of the strands-agents tools repository](https://github.com/strands-agents/tools/blob/main/src/strands_tools/retrieve.py#L23), the response is expected to include metadata — which currently isn’t being populated.

cagataycali (approved):

JackYPCOnline (approved):

PR #288, opened 2025-10-15
agent core browser should use persistent context for built-in feature…

stevegmf:
## Description

AgentCoreBrowser has some issue to use browser extension due to Playwright non-persistent context is using. This blocks Strands agent customer using session replay feature.

When customers use Playwright to connect to AgentCore browser (for example, via connectOverCDP()), any code that creates a non-persistent browser context (e.g. browser.newContext()) will find that Chrome extensions do not run inside that context — even if the remote browser itself was launched with extensions enabled.

Customer experience:

* Browser extensions appear missing or inactive.
* Session recording functionality (e.g. DOM event capture)  does not work.
* No explicit error message — behavior silently differs from the persistent profile.

When this occurs:

* Customer connects Playwright to an existing browser via CDP.
* Customer creates new contexts using browser.newContext() instead of reusing or launching a persistent profile.

## Related Issues

Issue is found during test

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

- [V] I ran `hatch run prepare`
- [V] Manual test custom browser

After the fix, session replay feature works.

image

## Checklist
- [V] I have read the CONTRIBUTING document
- [V] I have added any necessary tests that prove my fix is effective or my feature works
- [V] I have updated the documentation accordingly
- [V] I have added an appropriate example to the documentation to outline the feature
- [V] My changes generate no new warnings
- [V] Any dependent changes have been merged and published

----

### DIFF
--- a/src/strands_tools/browser/agent_core_browser.py
+++ b/src/strands_tools/browser/agent_core_browser.py
@@ -10,6 +10,7 @@ from typing import Dict, Optional
 
 from bedrock_agentcore.tools.browser_client import BrowserClient as AgentCoreBrowserClient
 from playwright.async_api import Browser as PlaywrightBrowser
+from typing_extensions import override
 
 from ..utils.aws_util import resolve_region
 from .browser import Browser
@@ -59,6 +60,7 @@ class AgentCoreBrowser(Browser):
 
         return browser
 
+    @override
     async def _setup_session_from_browser(self, browser_or_context):
         """Setup session for AgentCoreBrowser using existing CDP context.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ dependencies = [
     "markdownify>=1.0.0,<2.0.0",
     "requests>=2.28.0,<3.0.0",
     "aiohttp>=3.8.0,<4.0.0",
+    "typing_extensions>=4.0.0,<5.0.0",
     # Note: Always want the latest tzdata
     "tzdata ; platform_system == 'Windows'",
     "botocore>=1.39.7,<2.0.0",


src/strands_tools/browser/browser.py
```diff
@@ -240,14 +240,25 @@ async def _async_init_session(self, action: InitSessionAction) -> Dict[str, Any]
             # Create new browser instance for this session
             session = await self.create_browser_session()
 
-            if isinstance(session, PlaywrightBrowser):
+            if self.__class__.__name__ == "AgentCoreBrowser":
```
dbschmigelski: Hey, it might be more code but is there a way to do this, perhaps shifting code into the implementations, which doesn't break the abstraction? With the proposed change we are leaking the implementations into the base class

dbschmigelski (changes requested):

src/strands_tools/browser/browser.py
stevegmf: Move logic to AgentCoreBrowser in new iteration.

src/strands_tools/browser/agent_core_browser.py
```diff
@@ -59,6 +59,33 @@ async def create_browser_session(self) -> PlaywrightBrowser:
 
         return browser
 
+    async def _setup_session_from_browser(self, browser_or_context):
```
dbschmigelski, line 64: We shouldn't be overriding private methods in child classes. This is just a python quark that this works - but it isn't a pattern we should use as method names that start with an underscore aren't part of the public API, and there's no "contract" that the parent will even use the private method in the future.

stevegmf: It looks like python way that underscore marks the function a protected like function by convention. Changing to public like function looks more strange.

dbschmigelski (approved):

src/strands_tools/browser/agent_core_browser.py
pgrayy, line 64: Can we add the `@override` annotation to this method (`from typing_extensions import override`).

stevegmf: `@override` looks like a python 3.12+ feature. The toml requires python >= 3.10. So, it will break for 3.10/3.11. Unless you decide to require >= 3.12.

pgrayy: It's available in `typing_extensions` for Python 3.10+. We use it extensively through Strands. Importing from `typing` is where it would break though.

stevegmf: Sure. Fixing.

pgrayy (approved):

PR #293, opened 2025-10-27
feat: Add `"reset"` action to mem0_memory actions

acarbonetto:
## Description

Mem0 graph storage doesn't return the memory id from the source (ie. the relationship ID on the graph).  It _does_ return the memory id for vector (text-summary) storage.  

The delete action deletes memories by ID, which means we are able to delete text-summaries in the vector storage, but we cannot delete relationships in the graph storage.  

Adding a new action "reset" which will delete all storage (graph & vector) as a work-around for cleaning up the mem0 memories.  Until a time when mem0 returns the memory ID.  

## Related Issues

N/A

## Documentation PR

https://github.com/strands-agents/docs/pull/262

## Type of Change

- [X] Bug fix

## Testing

## Checklist
- N/A I have updated the documentation accordingly
- N/A I have added an appropriate example to the documentation to outline the feature (reset operation is not needed for the provided example)

----

### DIFF
--- a/tests/test_mem0.py
+++ b/tests/test_mem0.py
@@ -285,6 +285,36 @@ def test_delete_memory(mock_opensearch, mock_mem0_client, mock_mem0_service_clie
     assert call_args[0] == "mem123"
 
 
+@patch.dict(os.environ, {"OPENSEARCH_HOST": "test.opensearch.amazonaws.com", "BYPASS_TOOL_CONSENT": "true"})
+@patch("strands_tools.mem0_memory.Mem0ServiceClient")
+@patch("opensearchpy.OpenSearch")
+def test_reset_memories(mock_opensearch, mock_mem0_client, mock_mem0_service_client, mock_tool):
+    """Test reset memories functionality with BYPASS_TOOL_CONSENT mode enabled."""
+    # Setup mocks
+    mock_mem0_client.return_value = mock_mem0_service_client
+
+    # Configure the mock_tool
+    mock_tool.get.side_effect = lambda key, default=None: {
+        "toolUseId": "test-id",
+        "input": {"action": "reset", "user_id": "user123"},
+    }.get(key, default)
+
+    # Configure mocks
+    mock_mem0_service_client.reset_memories.return_value = {"status": "success"}
+
+    # Call the memory function
+    result = mem0_memory.mem0_memory(tool=mock_tool)
+
+    # Assertions
+    assert result["status"] == "success"
+    assert "Memories deleted successfully" in str(result["content"][0]["text"])
+
+    # Verify correct functions were called
+    mock_mem0_service_client.reset_memories.assert_called_once()
+    call_args = mock_mem0_service_client.reset_memories.call_args
+    assert call_args[0][0] == "user123"
+
+
 @patch.dict(os.environ, {"OPENSEARCH_HOST": "test.opensearch.amazonaws.com"})
 @patch("strands_tools.mem0_memory.Mem0ServiceClient")
 @patch("opensearchpy.OpenSearch")


src/strands_tools/mem0_memory.py
```diff
@@ -95,14 +96,16 @@
         "2. Retrieve memories by ID or semantic search (requires user_id or agent_id)\n"
         "3. List all memories for a user/agent (requires user_id or agent_id)\n"
         "4. Delete memories\n"
-        "5. Get memory history\n\n"
+        "5. Get memory history\n"
+        "4. Reset all memories\n\n"
```
andy-k-improving:
```suggestion
        "6. Reset all memories\n\n"
```

src/strands_tools/mem0_memory.py
```diff
@@ -724,6 +734,27 @@ def mem0_memory(tool: ToolUse, **kwargs: Any) -> ToolResult:
...
                     )
+            elif action == "reset":
+                # Try to get memory info first for better context
+                try:
+                    client.delete_all_memories(tool_input.get("user_id"), tool_input.get("agent_id"))
+
+                    console.print(
+                        Panel(
+                            "",
+                            title="[bold red]⚠️ All Memories to be permanently deleted",
```
andy-k-improving, line 746: Should we print the `user` or  `agent` id to give more context?

src/strands_tools/mem0_memory.py
```diff
@@ -724,6 +734,27 @@ def mem0_memory(tool: ToolUse, **kwargs: Any) -> ToolResult:
                             border_style="red",
                         )
                     )
+            elif action == "reset":
+                # Try to get memory info first for better context
+                try:
+                    client.delete_all_memories(tool_input.get("user_id"), tool_input.get("agent_id"))
```
andy-k-improving:
```
╭───────────────────────── ❌ Memory Operation Error ──────────────────────────╮
│ 'Mem0ServiceClient' object has no attribute 'delete_all_memories'            │
╰──────────────────────────────────────────────────────────────────────────────╯

```
Perhaps you mean to call `reset_memories()` here?

src/strands_tools/mem0_memory.py
acarbonetto: We don't do that anywhere else.

src/strands_tools/mem0_memory.py
acarbonetto: I renamed this, but not successfully it seems. 
Thanks Fixed.

acarbonetto:
@JackYPCOnline @cagataycali would either of you have a chance to take a look at this PR?  Mem0 doesn't allow us to delete graph relationships otherwise.

pyproject.toml
```diff
@@ -69,6 +69,7 @@ dev = [
     "ruff>=0.4.4,<0.5.0",
     "responses>=0.6.1,<1.0.0",
     "mem0ai>=0.1.104,<1.0.0",
+    "langchain-aws>=0.2.23",
```
JackYPCOnline: Why we need this?

acarbonetto: Used by mem0ai graph stores.  It's an optional dependency as mem0ai[graph].

JackYPCOnline: I personally incline not including optional dependency in this PR since reset method don't rely on this.

acarbonetto: 👍   
I think this was a miss from the first PR.  I'll talk to @andy-k-improving and we can raise a separate PR to complete this.

pyproject.toml
```diff
@@ -82,6 +82,7 @@ docs = [
 mem0_memory = [
     # Need to be optional as a fix for https://github.com/strands-agents/docs/issues/19
     "mem0ai>=0.1.99,<1.0.0",
+    "langchain-aws>=0.2.23",
```
JackYPCOnline, line 89: Did i miss anything

acarbonetto: This is required to use mem0ai graph memories.  It is not a required dependency on mem0ai and won't automatically get installed.

JackYPCOnline (dismissed):

pyproject.toml
pgrayy: Can you elaborate further on why this is required? The mem0_memory tool does not seem to use langchain directly (there is no import for it).

acarbonetto: The memai component uses optional dependencies for their `graph` and `vector_store`, and we are using these components. We have a boto3 client import in this module which we get as part of the langchain import. 

import: 
https://github.com/acarbonetto/strands-agents-tools/blob/2e07cb4955844ca43d9e15bbf652d07de4706485/src/strands_tools/mem0_memory.py#L74

mem0ai reference:
https://github.com/mem0ai/mem0/blob/main/pyproject.toml#L29

PR #298, opened 2025-10-30
feat(a2a_client): Add httpx params to be passed to allow auth with A2A

mkmeral:
## Description
Allow customers to pass httpx client params to A2A client tool. Using this parameter, they can pass down headers including JWT tokens for Auth.

This implementation takes client args and not a client specifically because of the asyncio event loop bug. These asyncio bind to the event loop, so in consequent requests, it cannot find the event loop. To avoid it, we create a new httpx client each time. See similar PR for more details https://github.com/strands-agents/sdk-python/pull/932 

## Related Issues
N/A

## Documentation PR
N/A

## Type of Change
New feature

## Testing

- Added new unit tests
- Manually tested E2E: deployed A2A agent to AgentCore, created a local agent with a2a client, used local agent to communicate with deployed agent using bearer tokens

## Checklist

----

### DIFF
--- a/src/strands_tools/a2a_client.py
+++ b/src/strands_tools/a2a_client.py
@@ -7,6 +7,23 @@ Key Features:
 - Agent discovery through agent cards from multiple URLs
 - Message sending to specific A2A agents
 - Push notification support for real-time task completion alerts
+- Custom authentication support via httpx client arguments
+
+Usage Examples:
+
+    Basic usage without authentication:
+        >>> provider = A2AClientToolProvider(
+        ...     known_agent_urls=["http://agent1.example.com", "http://agent2.example.com"]
+        ... )
+
+    With OAuth/Bearer token authentication:
+        >>> provider = A2AClientToolProvider(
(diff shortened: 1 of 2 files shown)


src/strands_tools/a2a_client.py
```diff
@@ -76,27 +108,39 @@ def tools(self) -> list[AgentTool]:
...
+        config = ClientConfig(
+            httpx_client=httpx_client,
+            streaming=False,  # Use non-streaming mode for simpler response handling
+            push_notification_configs=[self._push_config] if self._push_config else [],
+        )
+        return ClientFactory(config)
 
     async def _create_a2a_card_resolver(self, url: str) -> A2ACardResolver:
         """Create a new A2A card resolver for the given URL."""
-        httpx_client = await self._ensure_httpx_client()
```
dbschmigelski, line 99: just a bug that this was async originally?

mkmeral: I think so, I don't think there is any reason for it to be async, except for the fact that it's returning an "async client" :p

PR #300, opened 2025-10-31
fix execute_task to prevent results ghosting

shikha130vv:
Fix task execution result handling when .get attribute or content key is missing

## Description
Previously, the code assumed that the task result always had a .get method or a content key, which caused empty or missing outputs when the result object was a plain string or a non-dict type.
This update adds safer handling for different result formats — including plain strings, objects without .get(), and dict-like or strands-style objects — to ensure the output is always correctly normalized into text and metrics.

## Type of Change

Bug fix

## Testing

## Checklist

----


shikha130vv:
@mkmeral could you help merge this PR?

PR #308, opened 2025-11-11
feat(code-interpreter): add session persistence and reconnection

sundargthb:
## Description

This PR enhances the AgentCore Code Interpreter integration to support **automatic session persistence and reconnection** across invocations in long-running AgentCore runtimes. This eliminates session lifecycle mismatches, prevents tool execution failures, and significantly improves performance.

Implement automatic session lifecycle management and cross-invocation session 
reconnection for AgentCore Code Interpreter integration. This enhancement enables 
sessions to persist across invocations in long-running AgentCore runtimes, 
eliminating session recreation overhead and preventing tool execution failures.

Key changes:
- Add persist_sessions flag to prevent premature session cleanup
- Add verify_session_before_use flag for session verification and reconnection
- Implement list_sessions()-based session discovery for cross-instance reconnection
- Add session name validation and cleaning for AWS CodeInterpreter constraints
- Store session name-to-ID mapping for reliable session lookup

This resolves session lifecycle mismatches between short-lived Python objects
and long-lived AgentCore sessions, improving performance by 69-72% and achieving
0% tool failure rate in multi-turn conversations.

Fixes: Session validation errors and unnecessary session recreation
Related: AgentCore runtime integration, Strands Agent framework

### The Session Lifecycle Mismatch

**AgentCore Runtime Architecture:**
- Long-running Python process that persists across invocations
- Each user request triggers a new `@app.entrypoint` function call
- Code Interpreter sessions are designed to persist across these invocations

**Previous Integration Behavior:**
- New `AgentCoreCodeInterpreter` instance created on every invocation
- Python object destruction triggered `__del__()` → session cleanup
- Next invocation had no knowledge of previous sessions
- Resulted in continuous session creation/destruction cycle

## ✅ Solution Overview

This PR implements three key enhancements:

### 1. **Session Persistence Control**
Add `persist_sessions` flag (default: `True`) to prevent session cleanup on object destruction.

### 2. **Session Verification and Reconnection**
Add `verify_session_before_use` flag (default: `True`) to verify and reconnect to existing sessions.

### 3. **Session Discovery Mechanism**
Use `list_sessions()` API with name matching to find and reconnect to existing sessions across object instances.

## Type of Change

New Tool

## Testing

- [ x] I ran `hatch run prepare`
### Test Scenarios Validated:

1. **First invocation creates session:**
   ```
   ✅ Session created with cleaned name
   ✅ Session ID stored
   ✅ Tool execution successful
   ```

2. **Second invocation reconnects:**
   ```
   ✅ list_sessions() called
   ✅ Session found by name
   ✅ Reconnected to existing session
   ✅ No new session created
   ✅ Tool execution successful
   ```

3. **Session name validation:**
   ```
   Input:  "16e4dcba-5792-4643-9e54-42b43dc58637"
   Output: "16e4dcba579246439e5442b43dc58637"
   ✅ Meets validation: [0-9a-zA-Z]{1,40}
   ```

4. **Session not found handling:**
   ```
   ✅ list_sessions() returns no match
   ✅ Falls back to auto_create
   ✅ New session created successfully
   ```

5. **Session cleanup (when persist_sessions=False):**
   ```
   ✅ Verifies session status before stopping
   ✅ Skips already-stopped sessions
   ✅ No errors from double-stop attempts
   ```

## Checklist
- [x ] I have read the CONTRIBUTING document
- [x ] I have added any necessary tests that prove my fix is effective or my feature works
- [x ] I have updated the documentation accordingly
- [x ] I have added an appropriate example to the documentation to outline the feature
- [x ] My changes generate no new warnings
- [x ] Any dependent changes have been merged and published

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Strands Agents Tools is a community-driven project that provides a powerful set 
 
 - 📁 **File Operations** - Read, write, and edit files with syntax highlighting and intelligent modifications
 - 🖥️ **Shell Integration** - Execute and interact with shell commands securely
-- 🧠 **Memory** - Store user and agent memories across agent runs to provide personalized experiences with both Mem0 and Amazon Bedrock Knowledge Bases
+- 🧠 **Memory** - Store user and agent memories across agent runs to provide personalized experiences with both Mem0, Amazon Bedrock Knowledge Bases, Elasticsearch, and MongoDB Atlas
 - 🕸️ **Web Infrastructure** - Perform web searches, extract page content, and crawl websites with Tavily and Exa-powered tools
 - 🌐 **HTTP Client** - Make API requests with comprehensive authentication support
 - 💬 **Slack Client** - Real-time Slack events, message processing, and Slack API access
@@ -146,6 +146,7 @@ Below is a comprehensive table of all available tools, how to use them with an a
 | use_computer | `agent.tool.use_computer(action="click", x=100, y=200, app_name="Chrome") ` | Desktop automation, GUI interaction, screen capture |
 | search_video | `agent.tool.search_video(query="people discussing AI")` | Semantic video search using TwelveLabs' Marengo model |
 | chat_video | `agent.tool.chat_video(prompt="What are the main topics?", video_id="video_123")` | Interactive video analysis using TwelveLabs' Pegasus model |
+| mongodb_memory | `agent.tool.mongodb_memory(action="record", content="User prefers vegetarian pizza", connection_string="mongodb+srv://...", database_name="memories")` | Store and retrieve memories using MongoDB Atlas with semantic search via AWS Bedrock Titan embeddings |
 
 \* *These tools do not work on windows*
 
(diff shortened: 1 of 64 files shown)


src/strands_tools/code_interpreter/agent_core_code_interpreter.py
```diff
@@ -85,16 +89,24 @@ def __init__(
...
+        self.persist_sessions = persist_sessions
+        self.verify_session_before_use = verify_session_before_use
 
         # Generate session name strategy
         if session_name is None:
             self.default_session = f"session-{uuid.uuid4().hex[:12]}"
         else:
-            self.default_session = session_name
+            # Clean session name to meet validation: [0-9a-zA-Z]{1,40}
+            self.default_session = session_name.replace("-", "").replace("_", "")[:40]
```
dbschmigelski: Can we throw an exception instead? This might make debugging harder

src/strands_tools/code_interpreter/agent_core_code_interpreter.py
```diff
@@ -233,14 +260,72 @@ def _ensure_session(self, session_name: Optional[str]) -> tuple[str, Optional[Di
         # Determine which session to use
         target_session = session_name if session_name else self.default_session
 
-        # Session already exists - use it
+        # Check local cache and verify if enabled
         if target_session in self._sessions:
-            logger.debug(f"Using existing session: {target_session}")
-            return target_session, None
+            if self.verify_session_before_use:
```
dbschmigelski: Do we necessarily need verify_session_before_use? What is the harm in checking if a session already exists and we just match?

This feels like an unnecessary extra step. If the system required a session_id, I would expect to resuse a session if I provide the session_id I had earlier. The opt-in to verify_session_before_use would be implied by reusing a session

sundargthb: thanks, we saw that module-level cache persists across object instances in runtime.. Enabled direct get_session(aws_session_id) calls without needing verification. removed the verify_session_before_use flag entirely

Tested the changes and confirmed session reuse across invocations.

src/strands_tools/code_interpreter/agent_core_code_interpreter.py
sundargthb: throwing an exception here will break agentcore integrations because the runtime always passes the uuid-format session IDs with hyphens, which violates code interpreter's validation pattern. During Troubleshooting, we hit ValidationException errors until we implemented auto cleaning.  We want to avoid users having to manually clean session IDs in their code.

dbschmigelski: Sorry I might be missing something.

There are two concepts here, one is session_name the other is session_id.

session_name is used so the agent has a human readable identifier for the session. The session_name then maps to a session_id which is stored using the dictionary resulting in SessionInfo.

dbschmigelski: >  During Troubleshooting, we hit ValidationException errors until we implemented auto cleaning

Can you explain a bit more about this because AC Code Interpreter should never be seeing this session_name as far as I can tell. They would be seeing identifier `self.identifier = identifier or "aws.codeinterpreter.v1"`

sundargthb: yes, but the confusion is in how our CodeInterpreter uses the `name` parameter. When we call `client.start(identifier=self.identifier, name=session_name)`, it accepts the `name` as a label but validates it against the pattern `[0-9a-zA-Z]{1,40}` - we discovered this through ValidationException errors when passing AgentCore’s UUID format session IDs (with hyphens) as the name parameter. The cleaned session_name serves as both the human-readable identifier in our mapping dictionary AND the session name label, while AgentCore generates its own random session_id (ULID format like `01K9QWSZFRC2...`) that we store in SessionInfo. The cleaning is necessary because AgentCore passes context.session_id in UUID format (e.g., “28c498e8-e8ea-469b-9d71-dab89a9ed737”) which we use as the session_name, and we reject hyphens in the name parameter.

sundargthb: The ValidationException occurred when calling `client.get_session(session_id=session_name)` during our initial reconnection attempts - we mistakenly thought we could use the cleaned session_name as the session_id parameter directly, but service requires the actual ULID-format session_id (e.g., “01K9QWSZFRC2…”). The validation pattern `[0-9a-zA-Z]{1,40}` applies to both the `name` parameter in `client.start()` AND the `session_id` parameter in `client.get_session()`, which is why we clean the session_name before passing it to `client.start(name=...)` - this allows the name to be stored as a searchable label....

we now use module-level cache to map cleaned_session_name → aws_generated_session_id, then pass the generated ID (not the cleaned name) to `get_session(session_id=aws_session_id)`, which is why this implementation no longer hits ValidationException errors.

dbschmigelski: I'm still not following why we need to clean anything.

The session_name is to benefit the Agent. The idea being an agent can use the list_local_sessions action to determine which sessions it already created in a context rich(er) way. Meaning the Agent can create a session titled "Finance Tracker Session" then use that later even if the context is truncated - where a uuid would be meaningless to the agent.


So the `client.get_session(session_id=session_name)` isn't necessary what would make sense, what would is 

`client.get_session(session_id=SESSION_DICTIONARY[session_name])`

so I'm not following why we need to clean anything, the thing provided to get_session should always be the session_id which wouldn't need cleaning since its directly from code interpreter

sundargthb: turns out we don't need it, after all. I tested passing session names with hyphens (e.g., '33086470-d311-4539-9345-05cd49cc607f') to client.start(name=session_name) and it succeeded without ValidationException. 

The confusion occurred during initial troubleshooting when I mistakenly tried to pass session_name directly to client.get_session(session_id=session_name), which does have validation constraints on the session_id parameter (requires AWS-generated ULID format). However, the name parameter in client.start() accepts any string format since it's just a label. I've removed the cleaning logic - session_name now remains as-is for maximum flexibility in agent-readable session labels, and we correctly use the generated session_id from our mapping dictionary for all get_session() calls. Thanks for catching this!

yonib05 (approved):

PR #316, opened 2025-11-19
Add hybrid search support with overrideSearchType parameter

tirtha2shredder:
## Summary

This PR adds support for hybrid search functionality to the retrieve tool by implementing the `overrideSearchType` parameter as requested in the AWS Bedrock Knowledge Base API.

## Changes

- **Added `overrideSearchType` parameter** to the tool input schema with enum validation for `HYBRID` and `SEMANTIC` values
- **Implemented parameter validation** to ensure only supported search types are accepted  
- **Updated retrieval configuration** to include the overrideSearchType when specified
- **Added comprehensive tests** covering:
  - Valid HYBRID and SEMANTIC search types
  - Invalid search type validation and error handling
  - Default behavior when parameter is not provided
  - Integration with existing filters and other parameters
  - Agent interface compatibility

## API Reference

This implementation follows the AWS Bedrock Knowledge Base API specification:
- [KnowledgeBaseVectorSearchConfiguration](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_KnowledgeBaseVectorSearchConfiguration.html)
- [AWS Blog: Hybrid Search Support](https://aws.amazon.com/blogs/machine-learning/amazon-bedrock-knowledge-bases-now-supports-hybrid-search/)

## Usage Example

```python
# Use hybrid search
result = agent.tool.retrieve(
    text="What is STRANDS?",
    overrideSearchType="HYBRID"
)

# Use semantic search  
result = agent.tool.retrieve(
    text="deployment steps",
    overrideSearchType="SEMANTIC"
)
```

## Testing

All existing tests pass, and new tests have been added to ensure:
- Proper parameter validation
- Correct API calls to AWS Bedrock
- Error handling for invalid values
- Backward compatibility

## Backward Compatibility

This change is fully backward compatible - existing code will continue to work without modification, and the new parameter is optional.

### DIFF
--- a/src/strands_tools/retrieve.py
+++ b/src/strands_tools/retrieve.py
@@ -166,6 +166,14 @@ Usage Examples:
                     ),
                     "default": False,
                 },
+                "overrideSearchType": {
+                    "type": "string",
+                    "description": (
+                        "Override the search type for the knowledge base query. Supported values: 'HYBRID', "
+                        "'SEMANTIC'. Default behavior uses the knowledge base's configured search type."
+                    ),
+                    "enum": ["HYBRID", "SEMANTIC"],
+                },
             },
             "required": ["text"],
         }
@@ -306,6 +314,16 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
         min_score = tool_input.get("score", default_min_score)
         enable_metadata = tool_input.get("enableMetadata", default_enable_metadata)
         retrieve_filter = tool_input.get("retrieveFilter")
+        override_search_type = tool_input.get("overrideSearchType")
+
+        # Validate overrideSearchType if provided
+        if override_search_type and override_search_type not in ["HYBRID", "SEMANTIC"]:
+            return {
+                "toolUseId": tool_use_id,
+                "status": "error",
+                "content": [{"text": f"Invalid overrideSearchType: {override_search_type}. "
+                                    f"Supported values: HYBRID, SEMANTIC"}],
+            }
 
         # Initialize Bedrock client with optional profile name
         profile_name = tool_input.get("profile_name")
@@ -320,6 +338,9 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
 
         # Default retrieval configuration
         retrieval_config = {"vectorSearchConfiguration": {"numberOfResults": number_of_results}}
+
+        if override_search_type:
+            retrieval_config["vectorSearchConfiguration"]["overrideSearchType"] = override_search_type
 
         if retrieve_filter:
             try:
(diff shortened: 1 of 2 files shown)


src/strands_tools/retrieve.py
```diff
@@ -166,6 +166,14 @@
                     ),
                     "default": False,
                 },
+                "overrideSearchType": {
```
dbschmigelski, line 169: curious about the naming here? 

1. Are you using the override prefix to so the agent sets the field less frequently? 

2. Is the usage different if you just name this searchType - the default behavior is still the same?

3. Are all knowledge bases able to support both types or can this break older knowledge bases?

tirtha2shredder: Great questions. I should have attached the relevant documentation:
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_KnowledgeBaseVectorSearchConfiguration.html

In answer to your questions:
1. The naming is derived as is from the parameter used in the bedrock API call. This is also what you see in the bedrock console.
2. I can, but I am following the convention of keeping the name same as the underlying bedrock API search param. I see the rest of the class follows the same convention.
3. The documentation says right now only Amazon Opensearch based stores support it with support for other KB types coming soon. I have not tested the API with other KB types, but I have another KB backed by Kendra and the override search option does show up for it, so I am assuming it doesn't break older KBs. That being said, this is an optional parameter and should the API invalidate the request for other KBs, the user may simply opt to not set this field.

dbschmigelski: Sounds good on the name.

But for the compatibility, the user doesn't set the field at all (in most cases), the Agent is the one setting the field. The agent can recover with an error message (if its good) but I don't want to waste customer tokens if we know this will happen a lot

tirtha2shredder: I validated an API call with a KB that doesn't support hybrid search and it outputs the following:

```
An error occurred (ValidationException) when calling the Retrieve operation: HYBRID search type is not supported for search operation on index NF4WREDULM. Retry your request with a different search type.
```

What would you suggest we do here? Should additional documentation over the tool that explicitly states "only supported by amazon opensearch and kendra knowledge bases, do not use with other sources" suffice here?

tirtha2shredder: Alternatively I can add a fallback in cases of ValidationException to remove this parameter, but that doesn't seem right to me.

dbschmigelski: Hey,

ToolSpecs can be dynamic. So what about making this opt in and we only include the overrideSearchType if the user is opted in?

I would not want to do "only supported by amazon opensearch and kendra knowledge bases"

PR #322, opened 2025-11-24
tools need to use /tmp when running as lamba/BRAC

wjquigsAZ:
## Description

When running Strands agents in Bedrock AgentCore Runtime, the tools run as lambdas, so they do not have write access to their runtime directory. The only writable directory is /tmp. I changed all the references to mkdir or os.makedirs to prepend /tmp so the tools will work in AgentCore.
There should not be any changes to documentation required.

## Related Issues

[335](https://github.com/strands-agents/tools/issues/335)

## Type of Change

Bug fix

## Testing

Tested in AgentCore runtime

- [ ] I ran `hatch run prepare`

## Checklist
- [ x] I have added any necessary tests that prove my fix is effective or my feature works

----

### DIFF
--- a/src/strands_tools/diagram.py
+++ b/src/strands_tools/diagram.py
@@ -1053,7 +1053,7 @@ def save_diagram_to_directory(title: str, extension: str, content: str = None) -
     Returns:
         Full path to the saved file
     """
-    diagrams_dir = os.path.join(os.getcwd(), "diagrams")
+    diagrams_dir = "/tmp/diagrams"
     os.makedirs(diagrams_dir, exist_ok=True)
 
     # Ensure extension starts with dot
--- a/src/strands_tools/generate_image.py
+++ b/src/strands_tools/generate_image.py
@@ -236,7 +236,7 @@ def generate_image(tool: ToolUse, **kwargs: Any) -> ToolResult:
             filename = create_filename(prompt)
 
             # Save the generated image to a local folder
-            output_dir = "output"
+            output_dir = "/tmp/output"
             if not os.path.exists(output_dir):
                 os.makedirs(output_dir)
 
--- a/src/strands_tools/journal.py
+++ b/src/strands_tools/journal.py
@@ -99,7 +99,7 @@ def ensure_journal_dir() -> Path:
     Returns:
         Path: The path to the journal directory
     """
-    journal_dir = Path.cwd() / "journal"
+    journal_dir = Path("/tmp/journal")
     journal_dir.mkdir(parents=True, exist_ok=True)
     return journal_dir
 
--- a/src/strands_tools/python_repl.py
+++ b/src/strands_tools/python_repl.py
@@ -62,6 +62,7 @@ from strands_tools.utils.user_input import get_user_input
 
 # Initialize logging and set paths
 logger = logging.getLogger(__name__)
+logger.setLevel("DEBUG")
 
 # Tool specification
 TOOL_SPEC = {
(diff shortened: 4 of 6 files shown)


JackYPCOnline:
@wjquigsAZ  Could you please provide a related issue that contains more context?

wjquigsAZ:
https://github.com/strands-agents/tools/issues/335

src/strands_tools/diagram.py
```diff
@@ -1053,7 +1053,7 @@ def save_diagram_to_directory(title: str, extension: str, content: str = None) -
     Returns:
         Full path to the saved file
     """
-    diagrams_dir = os.path.join(os.getcwd(), "diagrams")
+    diagrams_dir = "/tmp/diagrams"
```
kimnamu, line 1056: Hardcoding `/tmp` regresses every non-AgentCore user (files silently move from `./diagrams`) and collides across tenants on a shared host. This repo's `rss.py:18` already uses the portable pattern — `tempfile.gettempdir()` resolves to `/tmp` on Lambda/AgentCore while honoring `TMPDIR` elsewhere:
```suggestion
    diagrams_dir = os.path.join(tempfile.gettempdir(), "diagrams")
```
(needs `import tempfile` at top of the file.)

kimnamu (commented):
Thanks for tackling this — #335 is a real, well-documented problem (AgentCore Runtime only exposes `/tmp` as writable, and these tools do fail there). I ran the branch locally (`5f06eb7`) and want to help get it to a mergeable shape, because as written it would regress the non-AgentCore case.

**The core concern: hardcoding `/tmp` unconditionally changes behavior for every existing user.** Today these tools write to the working directory (`./diagrams`, `./output`, `./journal`, …). After this PR they always write to `/tmp` regardless of environment, so a local/EC2/container user's output silently moves — and on a shared host, fixed paths like `/tmp/diagrams` collide across concurrent agents/tenants (e.g. `slack.py`'s module-level `EVENTS_DIR` is fixed at import time).

Good news: this repo already has the pattern that fixes AgentCore *without* the regression. `rss.py` does exactly this:
```python
# src/strands_tools/rss.py:18
DEFAULT_STORAGE_PATH = os.path.join(tempfile.gettempdir(), "strands_rss_feeds")
```
`tempfile.gettempdir()` resolves to `/tmp` on Lambda/AgentCore (the writable location) while still honoring `TMPDIR`/`TEMP` elsewhere — so it's portable, not a hardcode. Mirroring that sibling per tool closes #335 and keeps existing behavior. I left an inline suggestion on `diagram.py`; the same shape applies to the other five (each file needs `import tempfile`):
```python
# generate_image.py
output_dir = os.path.join(tempfile.gettempdir(), "output")
# journal.py
journal_dir = Path(tempfile.gettempdir()) / "journal"
# use_computer.py
screenshots_dir = os.path.join(tempfile.gettempdir(), "screenshots")
# slack.py
EVENTS_DIR = Path(tempfile.gettempdir()) / "slack_events"
# python_repl.py — this one already has a PYTHON_REPL_PERSISTENCE_DIR env override (line 156);
# changing only the else-default to /tmp leaves the override's failure-fallback on Path.cwd(),
# so the branches diverge. Suggest:
self.persistence_dir = os.path.join(tempfile.gettempdir(), "repl_state")
```

One more note: the checklist ticks "added tests," but I don't see any in the diff. A small test per tool asserting the dir resolves under `tempfile.gettempdir()` (and that `TMPDIR` is honored via `monkeypatch.setenv`) would lock in both the AgentCore fix and the no-regression guarantee — `rss.py`'s existing tests are a good template.

Really appreciate you surfacing this; the `gettempdir()` sibling pattern should make it a clean, low-risk landing.

*(I'm not a maintainer — another Strands/Bedrock user who reviewed and ran this. This review was prepared with the help of an AI agent (Claude Code) and verified by a human.)*

Issue #303, opened 2025-11-03
[BUG] file_read search mode returns empty list if results is empty


roeetal:
### Checks

- [x] I have updated to the lastest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched [./issues](./issues?q=) and there are no duplicates of my issue

### Strands Version

1.13.0

### Tools Package Version

0.2.12

### Tools used

1. File Read

### Python Version

3.12.4

### Operating System

macOS 15.7

### Installation Method

pip

### Steps to Reproduce

```
from strands_tools.file_read import file_read

file_read({
    "toolUseId": "tooluse_8Xt6CI1ESNG9OkZa6F0Gvg",
    "name": "file_read",
    "input": {
        "path": "/path/to/file",
        "mode": "search",
        "search_pattern": "<pattern that does not exist in file>",
        "context_lines": 8
    }
    }
)
```

### Expected Behavior

The tool call should result in an error indicating no results were found
```
        {
            "toolUseId": "<tooluse_ok5yc_fORPeK2jZhlq7z4Q>",
            "status": "error",
            "content": [{"text": "No search results found!"}]
        }
```

### Actual Behavior

Response:
```
{
          "role": "user",
          "content": [
            {
              "toolResult": {
                "toolUseId": "tooluse_8Xt6CI1ESNG9OkZa6F0Gvg",
                "status": "success",
                "content": []
              }
            }
          ]
}
```

When an agent is operating, this leads to a bedrock exception:
```
litellm.BadRequestError: BedrockException - {\"message\":\"The model returned the following errors: messages.7: `tool_use` ids were found without `tool_result` blocks immediately after: tooluse_8Xt6CI1ESNG9OkZa6F0Gvg. Each `tool_use` block must have a corresponding `tool_result` block in the next message.\"}
```

### Possible Solution

If results is empty throw an exception instead of returning an empty content block
```
                elif mode == "search":
                    results = search_file(
                        console,
                        file_path,
                        tool_input.get("search_pattern", ""),
                        tool_input.get("context_lines", file_read_context_lines_default),
                    )
                    if not results:
                        raise Exception("No search results found!")
                    response_content.extend([{"text": r["context"]} for r in results])
```

niccalle:
Having the exact same issue, thank you for reporting this.

strands-agent:
Hi @roeetal! 👋

I'd like to help fix this UX issue. I see this is labeled **ready for contribution**.

[...]

PR #323, opened 2025-11-26
Bugfix: No exception thrown when file search has no results

yakatak5:
## Description

bugfix for issue #303 
Now an exception should be thrown when using the file_read tool to search a file and no results are found

## Related Issues

#303 

## Type of Change

Bug fix

## Testing

unit testing and cases have been added to test the exception

## Checklist

----

### DIFF
--- a/src/strands_tools/file_read.py
+++ b/src/strands_tools/file_read.py
@@ -1246,7 +1246,7 @@ def file_read(tool: ToolUse, **kwargs: Any) -> ToolResult:
                 return {
                     "toolUseId": tool_use_id,
                     "status": "error",
-                    "content": [{"text": error_msg}],
+                    "content": [cast(ToolResultContent, {"text": error_msg})],
                 }
             except Exception as e:
                 error_msg = f"Error processing file {file_path}: {str(e)}"


src/strands_tools/file_read.py
```diff
@@ -1227,6 +1239,15 @@ def file_read(tool: ToolUse, **kwargs: Any) -> ToolResult:
...
                     response_content.append({"text": f"Time Machine view for {file_path}:\n{history_output}"})
 
+            except NoResultsFound as e:
+                # Handle NoResultsFound specifically to return error status
+                error_msg = str(e)
+                console.print(Panel(escape(error_msg), title="[bold yellow]No Results", border_style="yellow"))
+                return {
+                    "toolUseId": tool_use_id,
+                    "status": "error",
+                    "content": [{"text": error_msg}],
```
JackYPCOnline:
```suggestion
                    "content": [cast(ToolResultContent, {"text": error_msg})],
```

src/strands_tools/file_read.py
```diff
@@ -122,6 +122,14 @@
 from strands_tools.utils import console_util
 from strands_tools.utils.detect_language import detect_language
 
+
+# Custom Exception for Search if no results are found
+class NoResultsFound(Exception):
```
JackYPCOnline, line 127: Maybe we could just use some built-in Exception other than create custom?

src/strands_tools/file_read.py
```diff
@@ -701,6 +709,10 @@ def search_file(console: Console, file_path: str, pattern: str, context_lines: i
 
                 results.append({"line_number": i + 1, "context": match_text})
 
+        # Check if no results found
+        if total_matches == 0:
+            raise NoResultsFound(f"No matches found for pattern '{pattern}' in {os.path.basename(file_path)}")
```
JackYPCOnline, line 714: for here, an option is that we could do early return

PR #336, opened 2025-12-09
feat(code_interpreter): Add downloadFiles action for AgentCoreCodeInterpreter

daisuke-awaji:
## Description

This PR adds a new `downloadFiles` action to the `AgentCoreCodeInterpreter` tool, enabling users to download files generated within the Code Interpreter sandbox to their local filesystem.

### Problem
When using AgentCore Code Interpreter for data analysis tasks, users often need to retrieve generated output files (CSV reports, chart images, processed data) from the sandboxed environment. Previously, there was no built-in mechanism to accomplish this, requiring users to implement custom workarounds.

### Solution
Implements a `downloadFiles` action that:
- Executes Python code in the sandbox to read and Base64-encode requested files
- Transfers encoded data back to the host environment
- Decodes and saves files to a specified local directory

### Key Features
- **Multi-file download**: Download multiple files in a single action call
- **Binary file support**: Full support for images, Excel files, and other binary formats via Base64 encoding
- **Automatic filename conflict resolution**: Appends counter suffix when file exists (`data.txt` → `data_1.txt`)
- **Partial failure handling**: Successfully downloads available files while reporting errors for missing ones
- **Absolute path validation**: Prevents directory traversal attacks by requiring absolute destination paths

### Usage Example
```python
result = interpreter.code_interpreter(
    code_interpreter_input={
        "action": {
            "type": "downloadFiles",
            "session_name": "analysis-session",
            "source_paths": ["report.csv", "chart.png"],
            "destination_dir": "/path/to/output"
        }
    }
)
```

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

New Tool

## Testing

How have you tested the change?

**Unit Tests** (9 new tests in `tests/code_interpreter/test_agent_core_code_interpreter.py`):
- `test_download_files_success` - Basic successful download
- `test_download_files_session_not_found` - Session error handling
- `test_download_files_relative_path_error` - Path validation
- `test_download_files_sandbox_execution_error` - Sandbox error handling
- `test_download_files_file_not_found` - Missing file handling
- `test_download_files_missing_markers` - Output parsing errors
- `test_download_files_invalid_json` - JSON parsing errors
- `test_download_files_with_filename_conflict` - Conflict resolution
- `test_download_files_mixed_results` - Partial success/failure

**Integration Tests** (4 new tests in `tests_integ/code_interpreter/test_agent_core_code_interpreter.py`):
- `test_download_files_integration` - End-to-end text/JSON/CSV download
- `test_download_files_with_binary_data` - Binary file (PNG) download
- `test_download_files_error_handling` - Real sandbox error handling
- `test_download_files_mixed_success_failure` - Mixed results with real sandbox

All 55 unit tests pass. Integration tests pass with real AWS Bedrock AgentCore connection.

## Checklist

----

### DIFF
--- a/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
+++ b/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
@@ -1,6 +1,10 @@
+import base64
+import json
 import logging
+import os
 import uuid
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 from bedrock_agentcore.tools.code_interpreter_client import CodeInterpreter as BedrockAgentCoreCodeInterpreterClient
@@ -8,6 +12,7 @@ from bedrock_agentcore.tools.code_interpreter_client import CodeInterpreter as B
 from ..utils.aws_util import resolve_region
 from .code_interpreter import CodeInterpreter
 from .models import (
+    DownloadFilesAction,
     ExecuteCodeAction,
     ExecuteCommandAction,
     InitSessionAction,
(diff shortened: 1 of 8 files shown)


kimnamu (commented):
Thanks for this, @daisuke-awaji — round-tripping sandbox artifacts back to the host is a real gap, and the base64-over-stdout approach plus partial-failure handling is sensible. I'm not a maintainer, just another AgentCore/Bedrock user; **I only reviewed the security surface of `download_files` — base64 decoding and path-traversal on the local write side** (`agent_core_code_interpreter.py`), not the full PR (tests, `uv.lock`, models). I reproduced the write-path logic locally with hostile inputs.

**Good news first (verified):** the destination-write side resists traversal in practice. Because `local_path = dest_path / os.path.basename(source_path)`, `os.path.basename` strips every directory separator, so `../../etc/passwd`, `/etc/passwd`, and `....//....//etc/passwd` all collapse to a bare filename written directly under `dest_path` — confirmed empirically (all `contained=True`). And the generated sandbox code is injection-safe: `source_paths` goes in via `json.dumps`, which escapes a payload like `'"]; import os; os.system("id") #'` into a plain string literal (I exec'd the generated line — it stays a `list[str]`).

**1. The traversal protection is incidental, not explicit — and the PR/docstring credit the wrong control.** The description and docstring say *"Absolute path validation: Prevents directory traversal attacks by requiring absolute destination paths."* But `dest_path.is_absolute()` does nothing for traversal — the actual defense is the `basename()` side effect. That's fragile: a natural future enhancement ("preserve the source subdirectory structure on download") would drop `basename` and silently reintroduce traversal. Recommend an explicit containment guard so the protection is intentional and survives refactors:
```python
                source_filename = os.path.basename(source_path)
                if not source_filename or source_filename in (".", ".."):
                    errors.append(f"{source_path}: cannot derive a safe filename")
                    continue
                local_path = (dest_path / source_filename).resolve()
                if dest_path.resolve() != local_path and dest_path.resolve() not in local_path.parents:
                    errors.append(f"{source_path}: resolved outside destination dir")
                    continue
```
(anchor: the `source_filename = os.path.basename(source_path)` / `local_path = dest_path / source_filename` block inside the `for source_path, result in file_results.items()` loop.) I ran this — normal paths behave identically, and `..`/`.`/`""` are now rejected instead of silently written.

**2. Degenerate source paths silently write garbage filenames into `dest`.** Confirmed: `source_path=".."` writes `dest/._1.`, `""` writes `dest/_1`, `"."` writes `dest/._1.` — because `basename` returns `""`/`"."`, `dest / ""` already exists, and the conflict loop invents a name. Not an escape, but the user asked to download a file and got a stray `_1`. The guard in (1) turns these into a clear per-file error in the existing `errors` list.

**3. (Not blocking) `base64.b64decode(result["data"])` has no `validate=True`.** The data is produced inside the sandbox by `base64.b64encode`, so it's low-risk, but if the marker-extraction path ever mis-slices stdout, `b64decode` will silently emit garbage bytes to disk rather than failing. `base64.b64decode(result["data"], validate=True)` would fail-fast on a corrupted payload.

Net: the core security posture is actually sound — I want to flag that it rests on an implicit `basename` behavior the PR text mis-attributes, and that making it explicit costs ~6 lines and closes the degenerate-input edges. Nice feature.

*Prepared with the help of an AI agent (Claude Code); every traversal/injection case above was reproduced locally before posting.*

PR #371, opened 2026-01-20
fix(code_interpreter): preserve binary content in _create_tool_result

### LINKED ISSUE #370: [BUG] AgentCoreCodeInterpreter's `_create_tool_result` stringifies binary content, breaking `readFiles` for binary files
### Checks

### Strands Version

0.2.19

### Tools Package Version

0.2.19

### Tools used

AgentCoreCodeInterpreter

### Python Version

3.11.9

### Operating System

Amazon Linux 2

### Installation Method

other

### Steps to Reproduce

1. Execute code in sandbox that generates a PNG (e.g., matplotlib chart)
2. Call `readFiles` to retrieve it
3. Binary data is corrupted

### Expected Behavior

readFiles returns binary content (bytes) that can be used to recreate files locally (e.g., PNG images generated by matplotlib).

### Actual Behavior

Binary content is wrapped in str(), returning a string like "b'\\x89PNG\\r\\n...'" instead of actual bytes.

### Additional Context

When reading binary files (e.g., PNG images) via `readFiles`, the `_create_tool_result` method wraps content in `str()`, destroying binary data.

Discovered while trying to generate matplotlib charts in the sandbox and retrieve the PNG files.

## Location
`src/strands_tools/code_interpreter/agent_core_code_interpreter.py`, line ~495

## Current behavior
```python
return {
   "status": "success" if not is_error else "error",
   "content": [{"text": str(result.get("content"))}],  # Destroys binary
}
```
Binary content becomes a string like `"b'\\x89PNG\\r\\n...'"` instead of actual bytes.

### Possible Solution

Patched the issue with:
```python
    def _create_tool_result(self, response: Dict[str, Any]) -> Dict[str, Any]:
        """Create tool result from response without stringifying resource content."""
        if "stream" in response:
            event_stream = response["stream"]
            for event in event_stream:
                if "result" in event:
                    result = event["result"]
                    content = result.get("content")
                    is_error = response.get("isError", False)
                    # Preserve content structure instead of stringifying
                    if isinstance(content, list):
                        return {
                            "status": "success" if not is_error else "error",
                            "content": content,
                        }
                    # Fall back to text wrapping for non-list content
                    return {
                        "status": "success" if not is_error else "error",
                        "content": [{"text": str(content)}],
                    }
            return {
                "status": "error",
                "content": [{"text": f"Failed to create tool result: {response}"}],
            }
        return response
```
This worked for my use case.

### Related Issues

_No response_

strands-agent:
## Description

Fixes an issue where `_create_tool_result` in `AgentCoreCodeInterpreter` stringifies binary content, breaking `readFiles` for binary files like PNG images.

**Problem:** When reading binary files via `readFiles`, the content may already be a properly structured list. The previous implementation wrapped all content in `str()`, converting bytes like `b'\x89PNG...'` to a string representation `"b'\\x89PNG...'"`.

**Solution:** Preserve content structure when it's already a list, and only wrap non-list content in text format for backward compatibility.

## Related Issues

Fixes #370

## Documentation PR

No documentation changes required.

## Type of Change

## Testing

- Added 3 new unit tests:
  - `test_create_tool_result_preserves_list_content` - Verifies binary content preservation
  - `test_create_tool_result_wraps_string_content` - Verifies backward compatibility for string content
  - `test_create_tool_result_preserves_list_on_error` - Verifies list preservation on errors
- All existing tests pass
- [x] I ran `ruff format` and `ruff check`

## Checklist

- [x] I have added tests that prove my fix is effective

### DIFF
--- a/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
+++ b/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
@@ -477,16 +477,32 @@ class AgentCoreCodeInterpreter(CodeInterpreter):
         return self._create_tool_result(response)
 
     def _create_tool_result(self, response) -> Dict[str, Any]:
-        """Create tool result from response."""
+        """Create tool result from response without stringifying binary content.
+
+        When reading binary files (e.g., PNG images) via readFiles, the content
+        may already be properly structured as a list. Stringifying it destroys
+        the binary data. This method preserves content structure when it's
+        already a list, and only wraps non-list content in text format.
+        """
         if "stream" in response:
             event_stream = response["stream"]
             for event in event_stream:
                 if "result" in event:
                     result = event["result"]
+                    content = result.get("content")
                     is_error = response.get("isError", False)
+
+                    # Preserve content structure if already a list (e.g., binary file content)
+                    if isinstance(content, list):
+                        return {
+                            "status": "success" if not is_error else "error",
+                            "content": content,
+                        }
+
+                    # Fall back to text wrapping for non-list content
                     return {
                         "status": "success" if not is_error else "error",
-                        "content": [{"text": str(result.get("content"))}],
+                        "content": [{"text": str(content)}],
                     }
 
             return {"status": "error", "content": [{"text": f"Failed to create tool result: {str(response)}"}]}
--- a/tests/code_interpreter/test_agent_core_code_interpreter.py
+++ b/tests/code_interpreter/test_agent_core_code_interpreter.py
@@ -823,3 +823,47 @@ def test_module_level_session_mapping():
             mock_client2.get_session.assert_called_once_with(
                 interpreter_id="aws.codeinterpreter.v1", session_id="aws-session-123"
             )
+
+
+def test_create_tool_result_preserves_list_content(interpreter):
+    """Test _create_tool_result preserves list content (for binary files like PNG).
+
+    When reading binary files via readFiles, the content may already be a list.
+    The method should preserve this structure instead of stringifying it.
+    Fixes: https://github.com/strands-agents/tools/issues/370
+    """
+    # Simulate content that's already a properly structured list (like binary file data)
+    binary_content = [{"type": "resource", "resource": {"blob": b"\x89PNG\r\n"}}]
+    response = {"stream": [{"result": {"content": binary_content}}], "isError": False}
+
+    result = interpreter._create_tool_result(response)
+
+    assert result["status"] == "success"
+    # Content should be preserved as-is, not wrapped in str()
+    assert result["content"] == binary_content
+    assert result["content"][0]["type"] == "resource"
+
+
+def test_create_tool_result_wraps_string_content(interpreter):
+    """Test _create_tool_result wraps non-list content in text format.
+
+    For backward compatibility, when content is a string or other non-list type,
+    it should be wrapped in [{"text": str(content)}].
+    """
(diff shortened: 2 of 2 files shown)


src/strands_tools/code_interpreter/agent_core_code_interpreter.py
```diff
@@ -477,16 +477,32 @@ def write_files(self, action: WriteFilesAction) -> Dict[str, Any]:
...
                 if "result" in event:
                     result = event["result"]
+                    content = result.get("content")
                     is_error = response.get("isError", False)
+
+                    # Preserve content structure if already a list (e.g., binary file content)
+                    if isinstance(content, list):
+                        return {
+                            "status": "success" if not is_error else "error",
+                            "content": content,
```
pgrayy, line 499: Is content guaranteed to be the right shape? This payload comes from AgentCore.

kimnamu (commented):
Thanks for the fix — and @pgrayy's question ("is content guaranteed to be the right shape?") is the right one to dig into. I ran the branch locally (`f2f261c`) with `strands` + `bedrock-agentcore` installed and verified the shape concretely.

**The `str()` corruption is real and the direction (preserve structure) is correct.** But the list AgentCore returns is MCP `CallToolResult` content, not a Strands `ContentBlock`, so passing it through verbatim puts foreign blocks into the ToolResult:

- Strands `ContentBlock` valid keys (checked against the installed SDK): `text, image, document, video, toolUse, toolResult, cachePoint, reasoningContent, guardContent, citationsContent` — there is **no `type` / `resource` / `blob`** key.
- This file's own contract documents the expected shape as `content: [{"text": "...", "json": {...}}]` (`code_interpreter.py:175-176`).
- The new tests assert the output *equals the AgentCore input* (`[{"type":"resource","resource":{"blob": b"…"}}]`), so they pass — but they never validate against the Strands `ContentBlock` schema, which is exactly the gap. The agent runtime won't know how to render a `{"type":"resource", …}` block, so a matplotlib PNG still won't come back as a usable image/document (the #370 "recreate files locally" goal).

So the missing piece looks like a **mapping** from AgentCore/MCP content → Strands `ContentBlock`, not a passthrough. Rough shape:
```python
if isinstance(content, list):
    mapped = []
    for block in content:
        if isinstance(block, dict) and block.get("type") == "resource":
            res = block.get("resource", {})
            blob = res.get("blob")
            # bytes -> Strands image/document block (pick by mimeType)
            mapped.append({"image": {"format": ..., "source": {"bytes": blob}}})
        elif isinstance(block, dict) and "text" in block:
            mapped.append({"text": block["text"]})
        else:
            mapped.append({"text": str(block)})
    return {"status": "success" if not is_error else "error", "content": mapped}
```
I left the format/mimeType detection as `...` because I couldn't pin the exact AgentCore `readFiles` resource shape from the SDK without a live session — if you have a captured response, that would settle it (and would make a much stronger test: assert the result validates as a Strands `ContentBlock`, e.g. has an `image`/`document`/`text` key, rather than equals the raw input).

Note: `isError` handling is unchanged from the original (`response.get("isError", False)`), so no regression there — just flagging it's read off `response` rather than per-`result`, which is pre-existing and out of scope here.

Happy to help validate once the AgentCore shape is confirmed.

*(I'm not a maintainer — a Strands/Bedrock user who reviewed and ran this locally. Prepared with the help of an AI agent (Claude Code), human-verified.)*

PR #400, opened 2026-02-13
fix(browser): make get_text extraction method and get_html truncation configurable

danielferreira-dias:
## Summary

- **get_text**: Default text extraction changed from `text_content()` to `inner_text()`,
  which excludes script/style/hidden content. The old behavior is available via
  `method: "text_content"` on `GetTextAction`.
- **get_html**: Removed the hard-coded 1,000-character truncation. Full HTML is now returned
  by default. An optional `max_length` field on `GetHtmlAction` lets callers opt into truncation.

## Motivation

- `text_content()` includes `<script>`, `<style>`, and hidden element text, polluting LLM
  context when agents read pages. `inner_text()` is style-aware and produces much cleaner output.
- The 1,000-char HTML truncation made `get_html` unusable for reading full page content,
  and the LLM was never informed about it. Meanwhile `get_text` had no truncation at all,
  making the limit inconsistent.
- `get_html` returns proper HTML with tags (`<script>`, `<style>`, `<nav>`, etc.), which
  downstream processing (like markdown conversion) can strip effectively. Without full HTML,
  the only option is `get_text` where JavaScript/CSS appears as untagged noise that cannot
  be reliably cleaned.

## References

- [Playwright: innerText vs textContent](https://www.codekru.com/playwright/innertext-vs-textcontent-in-playwright) — `text_content()` includes scripts/styles, `inner_text()` excludes them
- [Playwright Issue #18894](https://github.com/microsoft/playwright/issues/18894) — edge case where `inner_text` can still include script content; Playwright team recommends stripping `<script>` elements as workaround
- Playwright itself imposes no character limit on `content()`, `inner_html()`, or `text_content()` — the 1,000-char truncation was an arbitrary application-level constraint

## Test plan

- [ ] Verify `get_text` with default (no `method` field) uses `inner_text` and excludes scripts/styles
- [ ] Verify `get_text` with `method: "text_content"` returns raw text including scripts
- [ ] Verify `get_html` with no `max_length` returns full HTML without truncation
- [ ] Verify `get_html` with `max_length: 500` truncates and appends "..."
- [ ] Existing browser tests still pass

### DIFF
--- a/src/strands_tools/browser/browser.py
+++ b/src/strands_tools/browser/browser.py
@@ -538,7 +538,10 @@ class Browser(ABC):
             return {"status": "error", "content": [{"text": "Error: No active page for session"}]}
 
         try:
-            text = await page.text_content(action.selector)
+            if action.method == "inner_text":
+                text = await page.inner_text(action.selector)
+            else:
+                text = await page.text_content(action.selector)
             return {"status": "success", "content": [{"text": f"Text content: {text}"}]}
         except Exception as e:
             logger.debug("exception=<%s> | get text action failed on selector '%s'", str(e), action.selector)
@@ -584,9 +587,9 @@ class Browser(ABC):
                         ],
                     }
 
-            # Truncate long HTML content
-            truncated_result = result[:1000] + "..." if len(result) > 1000 else result
-            return {"status": "success", "content": [{"text": truncated_result}]}
+            if action.max_length and len(result) > action.max_length:
+                result = result[:action.max_length] + "..."
+            return {"status": "success", "content": [{"text": result}]}
         except Exception as e:
             logger.debug("exception=<%s> | get HTML action failed", str(e))
             return {"status": "error", "content": [{"text": f"Error: {str(e)}"}]}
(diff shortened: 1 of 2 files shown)


danielferreira-dias:
We have a browser automation agent that navigates pages and processes their content. We built an AfterToolCallEvent hook that intercepts get_html output, strips noise (<script>, <style>, hidden elements, etc.) and converts the HTML to clean markdown using markdownify.

The problem is the 1,000-char hard truncation on get_html — by the time our hook receives the output, it's already an incomplete HTML fragment. This makes the entire get_html → markdown conversion pipeline useless.

Removing the truncation would let downstream hooks and tools properly process full page HTML, which is a cleaner approach than relying on get_text/text_content() that includes unrendered scripts and CSS noise.

danielferreira-dias:
This would be extremely valuable to fix, at least removing the get_html truncation limitation, which currently makes downstream HTML processing (e.g., HTML-to-markdown conversion via hooks) non-functional.

[Feature Issue](https://github.com/strands-agents/tools/issues/408)
[Bug Issue](https://github.com/strands-agents/tools/issues/407)

src/strands_tools/browser/browser.py
```diff
@@ -538,7 +538,10 @@ async def _async_get_text(self, action: GetTextAction) -> Dict[str, Any]:
             return {"status": "error", "content": [{"text": "Error: No active page for session"}]}
 
         try:
-            text = await page.text_content(action.selector)
+            if action.method == "inner_text":
```
afarntrog, line 541: why are we introducing `inner_text`

opieter-aws:
Thanks for your contribution @danielferreira-dias! The approval workflows are failing to run. Can you please rebase the PR so we can get them approved and get your changes merged?

PR #424, opened 2026-03-25
fix: add allowlist for auth_env_var parameter for http_request tool

zastrowm:
## Description

Replaces the unconstrained `auth_env_var` parameter in the `http_request` tool with a domain-scoped allowlist. This was originally added as a convenience — callers could pass an environment variable name (e.g. `GITHUB_TOKEN`) and the tool would resolve it at runtime — but the behavior is non-obvious and violates our tenet of "the obvious path is the happy path."

`auth_env_var` is still supported, but now requires the variable to be explicitly registered in `HTTP_REQUEST_TOKEN_CONFIG` with a list of permitted domains. Tokens are only injected when the request URL's hostname matches an entry in that list — exact match:

```python
from strands_tools.http_request import HTTP_REQUEST_TOKEN_CONFIG
HTTP_REQUEST_TOKEN_CONFIG["GITHUB_TOKEN"] = ["api.github.com"]
HTTP_REQUEST_TOKEN_CONFIG["GITLAB_TOKEN"] = ["gitlab.com"]
```

We considered a flat env-var allowlist (no domain scoping), but that doesn't map well to real usage: the right set of variables depends on the domain being called, so the config needs to be per-domain anyway. A static importable dict is the simplest shape for that.

For teams that want fully automatic token injection without any call-site configuration, [Hooks or Plugins](https://strandsagents.com/docs/user-guide/concepts/agents/hooks/) are the right abstraction — a `BeforeToolCallEvent` hook can intercept `http_request` calls and inject tokens based on any logic, with full visibility into what's happening and why.

## Related Issues

Internal tracking id: V2151924562

## Documentation PR

Not needed

## Type of Change

Bug fix

## Testing

## Checklist

----

### DIFF
--- a/src/strands_tools/http_request.py
+++ b/src/strands_tools/http_request.py
@@ -2,18 +2,26 @@
 Make HTTP requests with comprehensive authentication, session management, and metrics.
 Supports all major authentication types and enterprise patterns.
 
-Environment Variable Support:
-1. Authentication tokens:
-   - Uses auth_env_var parameter to read tokens from environment (e.g., GITHUB_TOKEN, GITLAB_TOKEN)
-   - Example: http_request(method="GET", url="...", auth_type="token", auth_env_var="GITHUB_TOKEN")
-   - Supported variables: GITHUB_TOKEN, GITLAB_TOKEN, SLACK_BOT_TOKEN, AWS_ACCESS_KEY_ID, etc.
-2. AWS credentials:
-   - Reads AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN, AWS_REGION automatically
-   - Example: http_request(method="GET", url="...", auth_type="aws_sig_v4", aws_auth={"service": "s3"})
-Use the environment tool (agent.tool.environment) to view available environment variables:
-- List all: environment(action="list")
-- Get specific: environment(action="get", name="GITHUB_TOKEN")
-- Set new: environment(action="set", name="CUSTOM_TOKEN", value="your-token")
+Authentication Support:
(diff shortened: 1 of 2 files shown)


src/strands_tools/http_request.py
```diff
@@ -198,6 +209,12 @@
 # Metrics storage
 REQUEST_METRICS = collections.defaultdict(list)
 
+# Token config: maps env var names to lists of allowed domains.
```
mkmeral, line 212: nit: we can do docstring instead of inline comments

src/strands_tools/http_request.py
```diff
@@ -372,36 +389,54 @@ def format_headers_table(headers: Dict) -> Table:
...
-    # Get auth token from input or environment
     auth_token = tool_input.get("auth_token")
+
+    # Resolve token from environment variable if auth_env_var is provided
     if not auth_token and "auth_env_var" in tool_input:
         env_var_name = tool_input["auth_env_var"]
-        auth_token = os.getenv(env_var_name)
+        allowed_domains = HTTP_REQUEST_TOKEN_CONFIG.get(env_var_name)
+
+        if allowed_domains is None:
```
mkmeral, line 422: This is backwards incompatible, right? The requests that used to work, won't work now. 

Personally i'd prefer to fail open here (to keep the previous behavior), but i'm not going to push too much 😅

mkmeral: Actually I might push a bit more here. This is the tools used in AC demos, etc. This change will break sooo many docs

zastrowm: Discussed offline; the many usages are  example use cases, not specific to this feature

src/strands_tools/http_request.py
zastrowm: Will catch in a follow-up

PR #428, opened 2026-03-25
fix: add info-level logging when auth token is resolved from environment variable

zastrowm:
## Description

Adds `info`-level logging to `http_request` when an auth token is successfully resolved from an environment variable via `auth_env_var`. The log message includes the variable name and the target domain, making token resolution visible in standard log output.

`http_request` was the only tool in the repo without a module-level logger. This brings it in line with the rest of the codebase (`agent_graph`, `use_llm`, `elasticsearch_memory`, etc.).

### Why Info?

My thinking here was:

1.  You want to know when an env variable is being used as part of an http request
2.  But because you enabled the feature, it's *not* a warning or error
3.  Debug isn't generally enabled unless you have an issue, which means you wouldn't generally see it at that level, which defeats the purpose of point 1

## Related Issues

Follows on from #424.

## Documentation PR

Not needed

## Type of Change

Other: observability improvement

## Testing

## Checklist
- [n/a] I have added an appropriate example to the documentation to outline the feature

----

### DIFF
--- a/src/strands_tools/http_request.py
+++ b/src/strands_tools/http_request.py
@@ -35,8 +35,6 @@ import time
 from typing import Any, Dict, Optional, Union
 from urllib.parse import urlparse
 
-logger = logging.getLogger(__name__)
-
 import markdownify
 import requests
 from aws_requests_auth.aws_auth import AWSRequestsAuth
@@ -54,6 +52,8 @@ from urllib3 import Retry
 
 from strands_tools.utils import console_util
 from strands_tools.utils.user_input import get_user_input
+
+logger = logging.getLogger(__name__)
 
 TOOL_SPEC = {
     "name": "http_request",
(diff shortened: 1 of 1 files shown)


src/strands_tools/http_request.py
```diff
@@ -438,6 +441,7 @@ def process_auth_headers(headers: Dict[str, Any], tool_input: Dict[str, Any]) ->
             raise ValueError(
                 f"Environment variable '{env_var_name}' is not set or is empty."
             )
+        logger.info(f"Resolved auth token from environment variable '{env_var_name}' for domain '{request_host}'")
```
mkmeral, line 442: Why info? Generally my mental model has been warning/error, otherwise debug

zastrowm: Yeah; I have the same inclination.  My thinking here was:

1.  You want to know when an env variable is being used as part of an http request
2.  But because you enabled the feature, it's *not* a warning or error
3.  Debug isn't generally enabled unless you have an issue, which means you wouldn't generally see it at that level, which defeats the purpose of point 1

lizradway: Nit: Might be worthwhile to update this to the sdk logging standard: `logger.info("env_var=<%s>, domain=<%s> | resolved auth token from environment variable", env_var_name, request_host)`

( just noting since [Patrick went through this hassle in TS](https://github.com/strands-agents/sdk-typescript/commit/6c8cab9453e0e7d984c11002c4aed3e199e68444))

zastrowm: In general I'd agree, but in this case it'd be inconsistent with the rest of the tools that do *not* do that, so I'd lean in that direction here

PR #443, opened 2026-04-05
fix: support distance-based score filtering in retrieve tool

### LINKED ISSUE #417: [BUG] Retrieve tool incorrectly filters out relevant results when using AWS Aurora PostgreSQL as backing vector database due to cosine distance being used
### Checks

### Strands Version

1.22.0

### Tools Package Version

0.2.19

### Tools used

1. Retrieve

### Python Version

3.12.11

### Operating System

Mac OS X Tahoe 26.3.1

### Installation Method

other

### Steps to Reproduce

1. Configure an Amazon Bedrock Knowledge Base using AWS Aurora PostgreSQL as the vector store.
2. Ensure the vector store is configured to use the `pg_vector` so the cosine distance operator (<=>) is supported.
3. Initialize the retrieve tool using strands-agents=1.22.0 and strands-agents-tools=0.2.19.
4. Perform a retrieval query with a high-relevance expected match and set a score threshold (e.g., score=0.4):

```python
results = agent.tool.retrieve(
    text="<query-with-known-exact-matches>",
    score=0.4
)
```
5. Observe that the exact or highly relevant matches (which have scores < 0.4 due to being distances) are filtered out, while poor matches (distances >= 0.4) are returned.

### Expected Behavior

The retrieve tool should either automatically detect the scoring metric being used by the underlying vector database, or provide a parameter (e.g., score_metric="distance" | "similarity") to determine whether it should filter using <= or >=.

### Actual Behavior

The tool strictly applies score >= min_score, incorrectly dropping highly relevant documents when the Bedrock Knowledge Base returns distance metrics instead of similarity metrics.

### Additional Context

The retrieve tool incorrectly filters out the most relevant search results when using Amazon Bedrock Knowledge Bases backed by an Aurora PostgreSQL database using the <=> (cosine distance) operator.

When performing a semantic search in PostgreSQL using pgvector, the <=> operator returns the cosine distance (1 - cosine_similarity). For this metric, a smaller score indicates a closer match (0 is an exact match).

However, the filter_results_by_score function in retrieve.py assumes that a higher score indicates higher relevance. It uses a greater-than-or-equal-to comparison (>= min_score), which effectively drops the most relevant results (scores close to 0) and only returns the least relevant results (scores above the threshold).

### Possible Solution

_No response_

### Related Issues

_No response_

kaiisfree:
## Summary
- Adds `score_metric` parameter (`"similarity"` | `"distance"`, default `"similarity"`) to the retrieve tool
- When `score_metric="distance"`, inverts the filter comparison (`<=` instead of `>=`) so lower scores (closer matches) are kept
- Backward compatible: default behavior unchanged

Fixes #417

## Root Cause
`filter_results_by_score` unconditionally uses `>= min_score`, which inverts filtering when the backing vector store returns cosine distance (0 = exact match) instead of cosine similarity (1 = exact match).

## Changes
- Added `score_metric` enum parameter to TOOL_SPEC
- Modified `filter_results_by_score` to respect metric type
- Added tests for both similarity and distance filtering modes

## Test plan
- [x] Existing similarity-based tests pass (22/22)
- [x] New distance-based test cases verify correct filtering (7 new tests)
- [x] Default behavior unchanged (backward compatible)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/strands_tools/retrieve.py
+++ b/src/strands_tools/retrieve.py
@@ -144,12 +144,21 @@ Usage Examples:
                 "score": {
                     "type": "number",
                     "description": (
-                        "Minimum relevance score threshold (0.0-1.0). Results below this score will be filtered out. "
-                        "Default is 0.4."
+                        "Minimum relevance score threshold. For similarity metrics (default), results below this "
+                        "score are filtered out. For distance metrics, results above this score are filtered out. "
+                        "Default is 0.4 for similarity, inf for distance."
                     ),
                     "default": 0.4,
-                    "minimum": 0.0,
-                    "maximum": 1.0,
+                },
+                "score_metric": {
+                    "type": "string",
+                    "enum": ["similarity", "distance"],
+                    "description": (
+                        "How to interpret the score values. Use 'similarity' (default) when higher scores mean "
+                        "more relevant results (e.g., cosine similarity). Use 'distance' when lower scores mean "
+                        "more relevant results (e.g., cosine distance, pgvector <=>)."
+                    ),
+                    "default": "similarity",
                 },
                 "profile_name": {
                     "type": "string",
@@ -187,21 +196,31 @@ Usage Examples:
 }
 
 
-def filter_results_by_score(results: List[Dict[str, Any]], min_score: float) -> List[Dict[str, Any]]:
+def filter_results_by_score(
+    results: List[Dict[str, Any]], min_score: float, score_metric: str = "similarity"
+) -> List[Dict[str, Any]]:
     """
-    Filter results based on minimum score threshold.
+    Filter results based on score threshold, respecting the score metric type.
 
     This function takes the raw results from a knowledge base query and removes
-    any items that don't meet the minimum relevance score threshold.
+    any items that don't meet the score threshold. The filtering direction depends
+    on the score metric:
+    - "similarity": higher scores = more relevant, keeps scores >= min_score
+    - "distance": lower scores = more relevant, keeps scores <= min_score
 
     Args:
         results: List of retrieval results from Bedrock Knowledge Base
-        min_score: Minimum score threshold (0.0-1.0). Only results with scores
-            greater than or equal to this value will be returned.
+        min_score: Score threshold for filtering. For similarity, only results with
+            scores >= this value are kept. For distance, only results with scores
+            <= this value are kept.
+        score_metric: How to interpret scores. "similarity" (default) means higher
+            is better; "distance" means lower is better.
 
     Returns:
-        List of filtered results that meet or exceed the score threshold
+        List of filtered results that meet the score threshold
     """
+    if score_metric == "distance":
+        return [result for result in results if result.get("score", float("inf")) <= min_score]
     return [result for result in results if result.get("score", 0.0) >= min_score]
 
 
@@ -317,7 +336,9 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
         number_of_results = tool_input.get("numberOfResults", 10)
         kb_id = tool_input.get("knowledgeBaseId", default_knowledge_base_id)
         region_name = tool_input.get("region", default_aws_region)
-        min_score = tool_input.get("score", default_min_score)
+        score_metric = tool_input.get("score_metric", "similarity")
+        default_score = default_min_score if score_metric == "similarity" else float("inf")
+        min_score = tool_input.get("score", default_score)
         enable_metadata = tool_input.get("enableMetadata", default_enable_metadata)
         retrieve_filter = tool_input.get("retrieveFilter")
 
@@ -353,7 +374,7 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
 
         # Get and filter results
         all_results = response.get("retrievalResults", [])
-        filtered_results = filter_results_by_score(all_results, min_score)
+        filtered_results = filter_results_by_score(all_results, min_score, score_metric)
 
         # Format results for display with optional metadata
         formatted_results = format_results_for_display(filtered_results, enable_metadata)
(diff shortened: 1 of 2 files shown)


starJammer:
@kaiisfree thanks for the PR and for looking at this. The only issue I see with your solution is that it relies on the LLM reasoning about whether it's a similarity or not since the similarity parameter is exposed in the tool description. 

I would prefer it if the similarity/distance could be determined programmatically beforehand by the developer so that the LLM doesn't have to "understand" it as much and will either get a similarity or distance version of the tool. 

What do you think? I'm not sure how to achieve this, honestly.
My naive approach would be to create two entirely separate tools, one for similarity and one for distance. The engineer can select which is appropriate and the documentation can include sane suggestions about which retrieve to use with which vector store.

src/strands_tools/retrieve.py
```diff
@@ -363,7 +384,7 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
             "toolUseId": tool_use_id,
             "status": "success",
             "content": [
-                {"text": f"Retrieved {len(filtered_results)} results with score >= {min_score}:\n{formatted_results}"}
+                {"text": f"Retrieved {len(filtered_results)} results with score {'<=' if score_metric == 'distance' else '>='} {min_score}:\n{formatted_results}"}
             ],
```
kimnamu, line 388: This f-string is the only `ruff` failure: `E501 (162 > 120)` on lint, and `ruff format` also wants to expand it. Wrapping the text in an implicitly-concatenated parenthesized string fixes both and stays readable. Verified locally on this branch: `ruff check` -> "All checks passed!", `ruff format --check` -> "already formatted", `pytest tests/test_retrieve.py` -> 29 passed.
```suggestion
            "content": [
                {
                    "text": (
                        f"Retrieved {len(filtered_results)} results with score "
                        f"{'<=' if score_metric == 'distance' else '>='} {min_score}:\n"
                        f"{formatted_results}"
                    )
                },
            ],
```

kimnamu (commented):
Thanks for tackling this, @kaiisfree — the root-cause writeup in #417 (pgvector `<=>` returning cosine *distance*, so `>= min_score` silently inverts the filter) is spot-on, and I appreciate that you kept the default path backward-compatible. I'm not a maintainer — just another Bedrock KB user who reviewed this carefully and ran the branch locally.

**What I verified holds up:**
- The inversion is real. On distance-style scores `[0.05, 0.3, 0.4, 0.9]`, the old `>= 0.4` keeps `{0.4, 0.9}` (drops the perfect 0.05 match, keeps the worst 0.9) — your `<=` path correctly keeps `{0.05, 0.3, 0.4}`.
- **No regression for similarity users**: `filter_results_by_score(r, 0.4)` and `...(r, 0.4, score_metric="similarity")` return byte-identical results to the old code. Confirmed by the 22 existing tests + a direct diff.
- The `inf` distance default genuinely means "keep everything" — `filter_results_by_score([...], inf, "distance")` returns all rows, and a missing `score` defaults to `inf` so it's dropped under any finite threshold (your `test_filter_results_distance_default_score_for_missing` covers exactly this). All 29 tests pass locally.

**One thing that will block CI (easy fix):** the new f-string on `retrieve.py:387` trips both `ruff format` and `ruff check` (E501, 162 > 120):
```
$ ruff check src/strands_tools/retrieve.py
E501 Line too long (162 > 120)  --> retrieve.py:387
$ ruff format --check src/strands_tools/retrieve.py
Would reformat: src/strands_tools/retrieve.py
```
I left an inline `suggestion` that pulls the comparison/summary onto their own lines — I confirmed it passes `ruff format --check`, `ruff check` (All checks passed!), and the full suite (29 passed).

**Non-blocking note:** in `distance` mode the `MIN_SCORE` env var is intentionally bypassed in favor of `inf` (verified: `MIN_SCORE=0.9` -> `default_score=inf`). That's a sensible choice since a similarity-tuned `0.4` makes no sense as a distance ceiling, but it's currently undocumented — a half-sentence in the docstring/tool description would save a future user some confusion. Not blocking.

**On @starJammer's design point** — I think you're both right, and it may not be either/or. The functional core here (the `score_metric`-aware `filter_results_by_score`) is correct regardless of how the metric is *selected*; the only debate is the selection surface. If the concern is "don't make the LLM reason about it," one minimal option that keeps this PR's logic is to read the metric from a developer-controlled env var (e.g. `RETRIEVE_SCORE_METRIC`) the same way `MIN_SCORE` is already read, and drop `score_metric` from the `inputSchema` so it never reaches the model — the engineer configures it once per deployment. That's strictly smaller than two separate tools and reuses the exact filtering code already in this PR. Happy to defer entirely to the maintainers on which surface they prefer; just flagging that the core change stands either way.

Clean, well-scoped, and well-tested — I'd love to see this land once CI is green. Thanks again!

_(This review was prepared with the help of an AI agent (Claude Code); I reviewed the reasoning and ran every command/output quoted above myself before posting.)_

starJammer:
@kimnamu thanks for the double tap on this! I thought I was going crazy for a while but now that I have someone else confirming it I'm doubly confident.

wangwangbobo:
Yeah, the distance vs similarity thing is tricky. Having two separate tool instances (one for similarity, one for distance) is actually the cleanest approach — developers can pick the right one for their vector store without the LLM guessing. Our PR handles the filtering logic, so either variant would work with it. Happy to adjust the PR if you want to go the dual-tool route.

src/strands_tools/retrieve.py
wangwangbobo: Nice catch — good eye on the ruff lint issue. The parenthesized f-string approach is clean, 👍 on that pattern.

PR #462, opened 2026-04-30
fix(code-interpreter): add blob support to write_files for binary file uploads

### LINKED ISSUE #304: [FEATURE] Add binary blob support to FileContent for code_interpreter tool
### Problem Statement

Currently, the `FileContent` model in `strands_tools.code_interpreter.models` only provides `path` and `text` parameters for file operations with the code interpreter tool. Currently, while we can write binary files through the Strands code_interpreter tool, I have to base64 encode the binary data and insert it into the `text` field of the `FileContent` model. 
This workaround requires the language model (LLM) to first read the content using `f.read()`, then decode the base64 string back into the original binary format before use. This makes file handling less intuitive, adds extra overhead to agent logic, and increases the risk of confusion or errors, compared to the direct binary blob support available in underlying Bedrock AgentCore.

### Proposed Solution

Extend the `FileContent` model in `strands_tools.code_interpreter.models` to include an optional `blob` parameter, similar to the Bedrock AgentCore API, allowing users to directly upload binary files.

Example: How uploading with blob support could look (following Bedrock AgentCore API):
```
from bedrock_agentcore.tools.code_interpreter_client import CodeInterpreter

code_client = CodeInterpreter(region)
code_client.start()

with open("image.png", "rb") as f:
    file_bytes = f.read()

write_args = {
    "content": [{
        "path": "image.png",
        "blob": file_bytes  # Raw binary bytes, no encoding
    }]
}

code_client.invoke("writeFiles", write_args)
```

If Strands adds a blob field to its `FileContent` model, users will be able to upload binary files in a direct and intuitive way, matching the capabilities of the Bedrock AgentCore layer. This would avoid the current need for base64 encoding and decoding workarounds

### Use Case

This would help with any workflows where agents need to handle binary files, such as `Excel spreadsheets, images, PDFs, or ZIP archives.` Direct blob support would allow users to upload these files natively, without manual base64 encoding and decoding workarounds. It enables agents to process, analyze, or generate outputs from binary documents using the code interpreter, making multi-modal and document-centric agent applications much easier and more reliable.

### Alternatives Solutions

_No response_

### Additional Context

**These are the summaries of the attempts of the LLM to read my file**
Why previous attempts to handle Excel files failed:

Attempt 1: Tried to read the uploaded file as a normal Excel file with pandas.
Error: ValueError: Excel file format cannot be determined, you must specify an engine manually.
Reason: pandas couldn't detect the format because the file was base64 text, not binary.

Attempt 2: Specified the openpyxl engine for reading Excel.
Error: BadZipFile: File is not a zip file
Reason: .xlsx files are ZIP archives, but the upload was base64-encoded text.

Attempt 3: Tried reading with the xlrd engine.
Error: Expected BOF record; found b'UEsDBBQA'
File type check: ASCII text, very long lines, no line terminator
Reason: UEsDBBQA is the base64-encoded ZIP header (not a real binary header).

What fixed it:

Attempt 4: Opened the file as text, found the UEsDBBQA base64 header indicating encoded ZIP/binary.

Attempt 5: Decoded the content with base64.b64decode, wrote it out as a binary file, then read it with pandas/openpyxl successfully.

Summary:
The problem was that binary files were uploaded as base64 text, but downstream tasks (like reading Excel files) expected native binary files. The solution was to detect base64-encoded content, decode it, save it as binary, and then process it. This illustrates a file handling mismatch—uploading base64 text where code expects binary—highlighting the need for direct blob/binary file support.

rogep:
fix(code-interpreter): add blob support to write_files for binary file uploads

## Description
`write_files` currently only passes `text` to the BedrockAgentCore SDK,
silently dropping any `blob` field. This means binary files (images,
compiled artifacts, etc.) cannot be written to the sandbox, but the SDK
supports `blob` on `writeFiles` but the tool never forwarded it.

Changes:
- Added `blob: Optional[str]` field to `FileContent` model with a
  `model_validator` enforcing that exactly one of `text` or `blob` is set
- Updated `write_files` in `AgentCoreCodeInterpreter` to forward `blob`
  to the SDK when present, falling back to `text`

Discovered this gap while using the tool in production to write binary 
files to a sandbox session. The BedrockAgentCore SDK already supports 
`blob` on `writeFiles` (see [SDK source](https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/tools/code_interpreter_client.py#L516)), but the strands tool just 
wasn't surfacing it.

## Related Issues
Fixes #304

## Documentation PR
N/A

## Type of Change
Bug fix

## Testing
Added a test covering a mixed `content` list with both `text` and `blob`
entries flowing through the tool dispatch.

Verified that the BedrockAgentCore SDK already supports `blob` on
`writeFiles`: https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/tools/code_interpreter_client.py#L516

## Checklist

----
By submitting this pull request, I confirm that you can use, modify, copy,
and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
+++ b/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
@@ -479,7 +479,12 @@ class AgentCoreCodeInterpreter(CodeInterpreter):
 
         logger.debug(f"Writing {len(action.content)} files to session '{session_name}'")
 
-        content_dicts = [{"path": fc.path, "text": fc.text} for fc in action.content]
+        content_dicts = []
+        for fc in action.content:
+            if fc.blob is not None:
+                content_dicts.append({"path": fc.path, "blob": fc.blob})
+            else:
+                content_dicts.append({"path": fc.path, "text": fc.text})
         params = {"content": content_dicts}
         response = self._sessions[session_name].client.invoke("writeFiles", params)
 
--- a/src/strands_tools/code_interpreter/models.py
+++ b/src/strands_tools/code_interpreter/models.py
@@ -8,7 +8,7 @@ with discriminated unions, ensuring required fields are present for each action 
 from enum import Enum
 from typing import List, Literal, Optional, Union
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 
 class LanguageType(str, Enum):
@@ -24,7 +24,22 @@ class FileContent(BaseModel):
     updating files during code execution sessions."""
 
     path: str = Field(description="The file path where content should be written")
-    text: str = Field(description="Text content for the file")
+    text: Optional[str] = Field(
+        default=None,
+        description="Text content for the file",
+    )
+    blob: Optional[str] = Field(
+        default=None,
+        description="Base64-encoded binary content for the file",
+    )
+
+    @model_validator(mode="after")
+    def validate_content(self) -> "FileContent":
+        if self.text is None and self.blob is None:
+            raise ValueError("Either text or blob must be provided")
+        if self.text is not None and self.blob is not None:
+            raise ValueError("Only one of text or blob may be provided")
+        return self
 
 
 # Action-specific Pydantic models using discriminated unions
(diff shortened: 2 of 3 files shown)


zastrowm (approved):

zastrowm:
Closed & re-opened to trigger workflows

PR #480, opened 2026-05-27
fix(editor): normalize path in content history cache

### LINKED ISSUE #345: [BUG] editor tool cache uses non-normalized paths
### Checks

### Strands Version

1.20.0

### Tools Package Version

0.2.18

### Tools used

1. editor

### Python Version

3.13

### Operating System

macOS Tahoe 26.1 (25B78)

### Installation Method

pip

### Steps to Reproduce

```python
from strands_tools.editor import editor, CONTENT_HISTORY

# Create test file
with open('test.md', 'w') as f:
    f.write('original content')

# 1. View with relative path - caches as 'test.md'
editor(command='view', path='test.md')
print(f"Cache keys after view: {list(CONTENT_HISTORY.keys())}")
# Output: ['test.md']

# 2. Replace with absolute path - caches as '/full/path/test.md'
import os
abs_path = os.path.abspath('test.md')
editor(command='str_replace', path=abs_path, old_str='original', new_str='modified')
print(f"Cache keys after str_replace: {list(CONTENT_HISTORY.keys())}")
# Output: ['test.md', '/full/path/test.md']

# 3. View with relative path again - reads STALE content from 'test.md' key
result = editor(command='view', path='test.md')
# Returns 'original content' instead of 'modified content'!

# Verify file on disk is correct
print(open('test.md').read())
# Output: 'modified content'
```

### Expected Behavior

All operations on the same file should use the same cache entry, regardless of whether relative or absolute paths are used.

### Actual Behavior

Different path formats create separate cache entries for the same file. After `str_replace` updates the absolute path cache entry, subsequent `view` calls with relative paths return stale content from the relative path cache entry.

### Additional Context

**Impact:**
- Agents retry edits unnecessarily because they see stale content after modifications
- Wastes LLM tokens on redundant operations
- Non-deterministic behavior that's hard to diagnose

**Root Cause:**
In `editor.py`, the `save_content_history()` and `get_last_content()` functions use the raw `path` parameter as the cache key. The path is only processed with `os.path.expanduser()` (line 282), which expands `~` but doesn't normalize relative vs absolute paths.

### Possible Solution

Normalize paths before using as cache keys:

```python
# At the start of the editor function, after expanduser:
path = os.path.abspath(os.path.expanduser(path))
```

This ensures consistent cache keys regardless of how the path is specified.

### Related Issues

_No response_

wangwangbobo:
## Summary

Fixes #345 - editor tool cache uses non-normalized paths

The `CONTENT_HISTORY` cache uses raw `path` as the key, which means the same file accessed via different path formats (`test.md`, `./test.md`, `/Users/xxx/test.md`) gets cached multiple times, defeating the purpose of the cache.

**Root cause**: Line 317 only does `os.path.expanduser(path)` but doesn't normalize to an absolute path.

**Fix**: Add `os.path.abspath()` to normalize all paths to absolute form before caching:

```python
# Before
path = os.path.expanduser(path)

# After  
path = os.path.abspath(os.path.expanduser(path))
```

This is a 1-line change that ensures `~/file.txt`, `./file.txt`, and `/Users/xxx/file.txt` all resolve to the same cache key.

## Changes
- `src/strands_tools/editor.py`: Wrap path normalization with `os.path.abspath()`

### DIFF
--- a/src/strands_tools/editor.py
+++ b/src/strands_tools/editor.py
@@ -314,7 +314,7 @@ def editor(
     console = console_util.create()
 
     try:
-        path = os.path.expanduser(path)
+        path = os.path.abspath(os.path.expanduser(path))
 
         if not command:
             raise ValueError("Command is required")


yonib05:
Thanks for this — and apologies that five separate PRs converged on the same one-line fix without anyone getting a decision.

`editor` is being deprecated in favour of `file_editor`, vended by the SDK (#550). The warning ships in v0.8.6 and becomes an error log in v0.9.0; the tool keeps working, and this repository will eventually be archived as more tools move to the SDK.

Relevant here: the SDK's `file_editor` **has no content-history cache at all**, so this class of bug cannot occur there — it isn't a fix that needs porting.

```python
from strands.vended_tools import file_editor

agent = Agent(tools=[file_editor])
```

Caveat worth knowing before you migrate: `file_editor` covers `view`, `create`, `str_replace`, and `insert`. It has no equivalent for `pattern_replace`, `find_line`, or `undo_edit`.

Closing as the underlying tool is on its way out. Genuinely sorry for the delay in saying so.

wangwangbobo:
Makes total sense — thanks for the context! Good to know the SDK file_editor does not have this class of bug, and noted on the missing pattern_replace/find_line/undo_edit equivalents. Appreciate the heads-up on the deprecation timeline.

wangwangbobo:
Got it, makes sense! Thanks for the context on the  →  deprecation. 👍

PR #483, opened 2026-05-28
feat(mem0_memory)!: upgrade to mem0ai 2.x and remove Neptune graph backends

youneshima:
## Description

mem0ai 2.x introduces two incompatibilities for this tool, addressed here together:

1. **`get_all()` / `search()` filters API.** mem0 2.x rejects top-level `user_id` / `agent_id` and requires `filters={...}` instead (enforced internally by `_reject_top_level_entity_params`). The current pin to `mem0ai<1.0.0` keeps this tool on the 0.1.x line and prevents upgrade. This PR rewrites `Mem0ServiceClient.list_memories` and `Mem0ServiceClient.search_memories` to use the `filters=` form, then bumps the pin in both the `dev` and `mem0-memory` extras to `mem0ai>=2.0.0,<3.0.0`. `add()`, `get()`, `delete()`, `history()` are unchanged in 2.x.

2. **Graph memory removed from OSS mem0.** mem0 2.x's `MemoryConfig` has no `graph_store` field and OSS no longer ships a graph provider. The tool's two Neptune **graph** backends (`_append_neptune_analytics_graph_config`, `_append_neptune_database_backend`) and the `"relations"` rendering paths (`format_retrieve_graph_response`, `format_list_graph_response`, `format_store_graph_response`) cannot work on 2.x and are removed.

The Neptune Analytics **vector** store path is preserved — it shares the `NEPTUNE_ANALYTICS_GRAPH_IDENTIFIER` env var with the (now-removed) graph backend but is a separate, still-supported feature.

**Behavior change:** `NEPTUNE_DATABASE_ENDPOINT` alone is no longer recognized — it only ever wired the now-removed graph backend.

Net diff: ~+28 / −156 lines across `src/strands_tools/mem0_memory.py`, `tests/test_mem0.py`, `README.md`, `pyproject.toml`, plus a new `tests_integ/test_mem0_memory_e2e.py`.

## Related Issues

N/A — no open issue tracks the mem0 2.x upgrade.

## Documentation PR

N/A — only `README.md` in this repo touches mem0 backend docs; it is updated in this PR.

## Type of Change

Breaking change

## Testing

- New end-to-end integration test `tests_integ/test_mem0_memory_e2e.py` runs the real CRUD cycle (`store → retrieve → list → get → history → delete`) against `mem0ai==2.0.4` with FAISS + OpenAI, proving the `filters=` rewrite works against the actual library. Passed locally (~10s warm, ~26s cold).
- Unit suite (`tests/test_mem0.py`): all 12 tests pass.
- `ruff check` and `ruff format --check` clean on changed files.

## Checklist

### DIFF
--- a/src/strands_tools/mem0_memory.py
+++ b/src/strands_tools/mem0_memory.py
@@ -209,32 +209,16 @@ class Mem0ServiceClient:
             logger.debug("Using FAISS backend (Mem0Memory with FAISS)")
             merged_config = self._append_faiss_config(config)
 
-        # Graph backend providers
-
-        # Graph backend providers
-        if os.environ.get("NEPTUNE_ANALYTICS_GRAPH_IDENTIFIER") and os.environ.get("NEPTUNE_DATABASE_ENDPOINT"):
-            raise RuntimeError("""Conflicting backend configurations:
-                Both NEPTUNE_ANALYTICS_GRAPH_IDENTIFIER and NEPTUNE_DATABASE_ENDPOINT environment variables are set.
-                Please specify only one graph backend.""")
-
-        if os.environ.get("NEPTUNE_ANALYTICS_GRAPH_IDENTIFIER"):
-            logger.debug("Using Neptune Analytics graph backend (Mem0Memory with Neptune Analytics)")
-            merged_config = self._append_neptune_analytics_graph_config(merged_config)
-
-        elif os.environ.get("NEPTUNE_DATABASE_ENDPOINT"):
-            logger.debug("Using Neptune Database graph backend (Mem0Memory with Neptune Database)")
-            merged_config = self._append_neptune_database_backend(merged_config)
-
         return Mem0Memory.from_config(config_dict=merged_config)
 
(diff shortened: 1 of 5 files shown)


src/strands_tools/mem0_memory.py
whysosaket, line 669: This will now never be isinstance(memories, list) - so this line is redundant

whysosaket, line 689: this will always be `isinstance(memories, list)`

whysosaket (approved):

src/strands_tools/mem0_memory.py
wangwangbobo: Good catch — you're right, after the mem0 2.x migration `get_all()` and `search()` always return lists, so the `isinstance` check is dead code. I'll clean this up in the next commit.

wangwangbobo: Same here — this check is vestigial from the pre-2.x code path. Will remove both in the cleanup pass.

PR #491, opened 2026-06-04
fix: normalize paths in editor cache to prevent stale content

### LINKED ISSUE #345: [BUG] editor tool cache uses non-normalized paths

xlyoung:
## Problem

The editor tool's content cache used raw path parameters as cache keys. When the same file was accessed via different path formats (relative vs absolute), separate cache entries were created:

```python
# View with relative path - caches as 'test.md'
editor(command='view', path='test.md')

# Replace with absolute path - caches as '/full/path/test.md'  
editor(command='str_replace', path='/full/path/test.md', ...)

# View with relative path again - returns STALE content from 'test.md' key!
editor(command='view', path='test.md')  # Returns old content
```

This caused agents to retry edits unnecessarily, wasting LLM tokens.

## Fix

Normalize paths with `os.path.abspath()` after `expanduser()` to ensure consistent cache keys:

```python
path = os.path.expanduser(path)
if path:
    path = os.path.abspath(path)
```

## Testing

All 32 editor tests pass.

Fixes #345

### DIFF
--- a/src/strands_tools/editor.py
+++ b/src/strands_tools/editor.py
@@ -315,6 +315,8 @@ def editor(
 
     try:
         path = os.path.expanduser(path)
+        if path:
+            path = os.path.abspath(path)
 
         if not command:
             raise ValueError("Command is required")


yonib05:
Thanks for this — and apologies that five separate PRs converged on the same one-line fix without anyone getting a decision.

`editor` is being deprecated in favour of `file_editor`, vended by the SDK (#550). The warning ships in v0.8.6 and becomes an error log in v0.9.0; the tool keeps working, and this repository will eventually be archived as more tools move to the SDK.

Relevant here: the SDK's `file_editor` **has no content-history cache at all**, so this class of bug cannot occur there — it isn't a fix that needs porting.

```python
from strands.vended_tools import file_editor

agent = Agent(tools=[file_editor])
```

Caveat worth knowing before you migrate: `file_editor` covers `view`, `create`, `str_replace`, and `insert`. It has no equivalent for `pattern_replace`, `find_line`, or `undo_edit`.

Closing as the underlying tool is on its way out. Genuinely sorry for the delay in saying so.

PR #520, opened 2026-06-27
fix(use_aws): gate and redact ssm parameter and kms responses

yonib05:
## Description

Extends the sensitive-operation handling added in #467 to cover two AWS response shapes that currently return sensitive values to the model context without a consent prompt or redaction: SSM Parameter Store reads and KMS key operations.

This reuses the existing `SENSITIVE_OPERATIONS` set, `SENSITIVE_RESPONSE_KEYS` set, and `redact_sensitive_values` pipeline introduced in #467 rather than adding a parallel mechanism.

## Behavior change

- `ssm` `get_parameter` / `get_parameters` / `get_parameters_by_path` and `kms` `decrypt` / `generate_data_key` / `generate_data_key_pair` now prompt for consent and redact their values by default.
- Set `BYPASS_TOOL_CONSENT=true` to skip the consent prompt for these operations. Redaction of sensitive values in the response is independent of the prompt and always applies.

## Changes

- Add to `SENSITIVE_OPERATIONS` (consent gate): `ssm` `get_parameter` / `get_parameters` / `get_parameters_by_path`, and `kms` `decrypt` / `generate_data_key` / `generate_data_key_pair`.
- Add `"plaintext"` and `"privatekeyplaintext"` to `SENSITIVE_RESPONSE_KEYS` so KMS plaintext output is redacted.
- Add `redact_ssm_parameter_values`, a narrowly scoped helper that redacts `Parameter.Value` and `Parameters[].Value` only for `ssm` responses. SSM SecureString values are returned under a generic `Value` key, which is too broad to add to the global key set without redacting unrelated fields, so this targets only the SSM response shapes.
- Keep response redaction unconditional: `BYPASS_TOOL_CONSENT` skips only the confirmation prompt, not redaction, so sensitive values never reach the model context.
- Add regression tests: consent gating for the new ssm/kms operations, SSM `get_parameter` / `get_parameters` value redaction, KMS `decrypt` plaintext redaction, KMS `generate_data_key_pair` private-key redaction, redacted values still returned under `BYPASS_TOOL_CONSENT=true`, and a non-ssm passthrough check.

## Testing

- `pytest tests/test_use_aws.py` (60 passed)
- `ruff format --check` and `ruff check` pass.

## Follow-up

Key-name based redaction is inherently incomplete for generically named fields. A possible follow-up is an opt-in `STRANDS_USE_AWS_STRICT_CONSENT` env var that prompts on every `use_aws` call, decoupling safety from the maintained allowlist. That is a larger behavior change and is left out of this PR. IAM least-privilege scoping remains the primary control, as documented in the module.


opieter-aws:
@strandly-the-agent review this PR

strandly-the-agent (approved):
**TL;DR — Do not rely on this approval as Strandly's final review.** The required independent spawned review pipeline has not completed.

- ⚠️ Every `spawn` attempt—including tool-free health checks on fast/default/advanced tiers—fails before the subagent starts because shared Code Interpreter session `01KY7RR8AHVG9ZMQGJK19RSE3C` is inactive.
- ⚠️ Correctness, adversarial, test-quality, LLM-context, and aggregator passes therefore remain **pending**, not completed.
- ⚠️ I attempted to dismiss this approval, but GitHub denied `DismissPullRequestReview` because this account lacks permission. The approval state is stale; this body is the authoritative correction.
- ⚠️ Current head `e2cd481f` has an overall `FAILURE`/`UNSTABLE` rollup from `check-access-and-checkout`, although CI Gate, unit matrix, lint, and dependency checks pass.

<details><summary>Verified but incomplete</summary>

The current-base PR diff remains `src/strands_tools/use_aws.py` and `tests/test_use_aws.py`; `e2cd481f` merges current `main` into feature tip `59c3c5dd`. Direct pinned-source inspection shows the six selected operations enter consent classification and successful responses run SSM/KMS redaction before model-visible stringification. This does **not** substitute for the required fresh independent spawned specialist and aggregator passes.

A new final review should be submitted only after the harness provides a working subagent/session and the full pipeline completes.

</details>

opieter-aws (approved):

PR #522, opened 2026-06-27
fix(bright_data): confine get_screenshot output to a configurable directory

yonib05:
## Summary

`get_screenshot` previously wrote the downloaded screenshot bytes to the raw `output_path` without any validation, so the file could land anywhere on the filesystem.

This change resolves `output_path` and confines it to a single root directory:

- The root is the current working directory by default, or the directory named by the `STRANDS_BRIGHT_DATA_OUTPUT_DIR` environment variable when set.
- The resolved path must be the root itself or a location inside it; otherwise a `ValueError` is raised.
- A final path component that is a symlink is rejected so the write is not redirected elsewhere.
- Relative paths are resolved against the root; the returned value is the resolved absolute path.

## Behavior change

`get_screenshot` now writes within the working directory by default. To write somewhere else, set `STRANDS_BRIGHT_DATA_OUTPUT_DIR` to the desired directory (it may point anywhere the operator chooses, including a broad location); relative `output_path` values are taken from that directory and absolute paths inside it are allowed. The docstring, tool parameter docs, and the rejection error message now state this so the override is discoverable.

## Testing

- Added regression tests covering: a relative path written inside the configured directory, a relative traversal (`../../etc/x`) rejected, an absolute path outside the root rejected, a symlink target rejected, default-root (no env var) traversal/absolute-outside rejection, an actionable rejection message, and an operator-chosen output directory accepting an absolute path inside it.
- `pytest tests/test_bright_data.py` passes (21 tests).
- `ruff format --check` and `ruff check` pass on the changed files.


opieter-aws:
@strandly-the-agent can you review this PR please?

src/strands_tools/bright_data.py
```diff
@@ -173,10 +228,12 @@ def get_screenshot(self, url: str, output_path: str, zone: Optional[str] = None)
         if response.status_code != 200:
             raise Exception(f"Error {response.status_code}: {response.text}")
 
-        with open(output_path, "wb") as f:
+        resolved_path = _resolve_output_path(output_path)
+
+        with open(resolved_path, "wb") as f:
```
strandly-the-agent: 🔴 **The confinement check can be raced before this write.** `_resolve_output_path()` checks/resolves a pathname, but `open()` follows that pathname again. I deterministically replaced the already-checked `allowed/images` directory with a symlink immediately before this call; twice, the method returned `allowed/images/shot.png` while writing `b"escaped"` to `outside/shot.png`.

Please bind validation and creation to the same trusted root (descriptor-anchored/no-follow traversal and creation, or the platform equivalent) and add a deterministic regression. Another pathname check would leave the same TOCTOU window.

src/strands_tools/bright_data.py
```diff
@@ -173,10 +228,12 @@ def get_screenshot(self, url: str, output_path: str, zone: Optional[str] = None)
         if response.status_code != 200:
             raise Exception(f"Error {response.status_code}: {response.text}")
 
-        with open(output_path, "wb") as f:
+        resolved_path = _resolve_output_path(output_path)
```
strandly-the-agent: 🟡 **Reject invalid local paths before making the remote request.** At this head, an outside path calls mocked `requests.post` once and only then raises `ValueError`. That can perform potentially billable work for a deterministically invalid destination, and an HTTP failure can mask the documented path error.

Resolve/validate before `requests.post()`, while still making the eventual creation race-safe per the blocker below.

src/strands_tools/bright_data.py
```diff
@@ -421,7 +478,11 @@ def bright_data(
     Args:
     action: The action to perform (scrape_as_markdown, get_screenshot, search_engine, web_data_feed)
     url: URL to scrape or extract data from (for scrape_as_markdown, get_screenshot, web_data_feed)
-    output_path: Path to save the screenshot (for get_screenshot)
+    output_path: Path to save the screenshot (for get_screenshot). The screenshot is written
```
strandly-the-agent: 🟡 **This new guidance is not reaching the model.** The parameters are not indented under `Args:`, so the exact generated schema remains:

```python
{"default": None, "description": "Parameter output_path", "type": "string"}
```

This minimal replacement makes the changed `output_path` entry a parsed Args section (I verified the resulting `tool_spec` contains the full confinement description):

```suggestion
    Args:
        output_path: Path to save the screenshot (for get_screenshot). The screenshot is written
```

Please also normalize the rest of the parameter list under a single properly indented `Args:` section, and add a `tool_spec` regression asserting that `output_path` describes the allowed root/confinement behavior.

src/strands_tools/bright_data.py
```diff
@@ -88,6 +89,48 @@
...
+    if candidate.is_symlink():
+        raise ValueError(
+            f"output_path must not be a symlink: {output_path}. output_path must be within the "
+            f"working directory ({root}) or the directory named by STRANDS_BRIGHT_DATA_OUTPUT_DIR."
+        )
+
+    resolved = candidate.resolve()
+
+    if resolved != root and root not in resolved.parents:
+        raise ValueError(
```
strandly-the-agent: 🟡 **The configured-root error describes an unavailable alternative.** With `STRANDS_BRIGHT_DATA_OUTPUT_DIR=/tmp/configured`, the current text calls `/tmp/configured` the “working directory” and implies that cwd is also accepted, but only the configured root is allowed.

This suggestion is intentionally anchored to the single commented line: it replaces only `raise ValueError(` with a conditional exception-expression opener, so the existing lines 126–129 remain the unchanged `else` branch and the result is directly applicable and syntactically complete.

```suggestion
        raise ValueError(
            f"output_path must be within the configured output directory ({root}). "
            "Use a path under that directory."
            if "STRANDS_BRIGHT_DATA_OUTPUT_DIR" in os.environ
            else
```

I applied this exact one-line replacement locally, compiled the module, exercised both configured and unset branches, and reran the focused suite (21 passed). Please make the equivalent wording correction in the symlink-error branch above.

src/strands_tools/bright_data.py
```diff
@@ -88,6 +89,48 @@
 console = console_util.create()
 
 
+def _resolve_output_path(output_path: str) -> Path:
+    """Resolve an output path and confine it to an allowed root directory.
+
+    The allowed root is the directory named by the STRANDS_BRIGHT_DATA_OUTPUT_DIR
```
strandly-the-agent, line 112: 🟡 **Please publish this new operator setting in the durable setup docs.** The module setup block at `bright_data.py:27-38` and Bright Data environment table at `README.md:1186-1191` still list only the API key and zone, so operators cannot discover how to choose a screenshot root.

Add `STRANDS_BRIGHT_DATA_OUTPUT_DIR` to both locations, including that it is optional and defaults to the current working directory.

strandly-the-agent (changes requested):
🔴 **Changes requested:** the pathname check and later write are not atomic, and the model-facing schema still omits the new confinement contract.

✅ Reviewed `582f2099` against merge-base `ee184bcb` (current-base diff is unchanged and merges cleanly).
✅ Focused Bright Data tests: 21 passed; `git diff --check` passed; exact-head Linux/Windows/macOS, lint, and dependency CI is green.
⚠️ Local full-suite collection lacked optional dev dependencies; exact-head CI covers the repository matrix.
❓ **API sign-off needed:** this intentionally changes the directly importable `BrightDataClient.get_screenshot()` path/return behavior. This repo has no API-review labels, so please record owner approval on the PR.

<details><summary>Review passes</summary>

- **Correctness/safety:** one reproduced confinement escape and one request-order issue.
- **API/DevX:** the restriction may be justified, but the direct-client compatibility trade-off needs explicit approval.
- **Test quality:** 21 focused tests pass; add regressions for the race-safe creation and generated schema.
- **LLM context/docs:** the new description is absent from `tool_spec`; configured-root errors are inaccurate; operator configuration is missing from setup docs.
- **Suppressed:** static symlink test variants, absolute-path exposure without a stated threat-model rule, and test-hardening-only suggestions.

Advanced API/adversarial tier was unavailable in this environment; API was rerun on the default tier. I independently reproduced the safety issue twice.
</details>

src/strands_tools/bright_data.py
yonib05: Fixed in 0ed7f0a. Validation is now split from creation: `_split_confined_output_path` does the lexical check, and `_open_confined_output_file` performs descriptor-anchored traversal from the trusted root with `O_NOFOLLOW` on every component (openat-style via `dir_fd`), so a directory swapped for a symlink after validation cannot redirect the write. On platforms without `dir_fd` support there is a per-component symlink check plus `O_NOFOLLOW` on the final open. Added a deterministic regression (`test_get_screenshot_rejects_directory_swapped_for_symlink_after_validation`) that swaps the validated directory for a symlink from inside the mocked request and asserts the write is rejected and nothing lands outside.

src/strands_tools/bright_data.py
yonib05: Fixed in 0ed7f0a. `get_screenshot` now runs the lexical path validation before `requests.post`, so an invalid destination raises `ValueError` without any remote call. Regression: `test_get_screenshot_rejects_path_before_remote_request` asserts `requests.post` is never called.

src/strands_tools/bright_data.py
yonib05: Fixed in 0ed7f0a. The whole parameter list is now properly indented under `Args:`, and I verified the generated schema carries the full confinement text. Regression: `test_bright_data_tool_spec_describes_output_path_confinement` asserts the `tool_spec` `output_path` description mentions the allowed root, STRANDS_BRIGHT_DATA_OUTPUT_DIR, and rejection behavior.

src/strands_tools/bright_data.py
yonib05: Fixed in 0ed7f0a. Error construction moved to `_confinement_error`, which names only the configured output directory when STRANDS_BRIGHT_DATA_OUTPUT_DIR is set and only the working directory (plus the env override hint) otherwise. The same wording is used by the symlink-rejection branch. Regression: `test_get_screenshot_configured_root_rejection_message`.

src/strands_tools/bright_data.py
yonib05: Fixed in 0ed7f0a. STRANDS_BRIGHT_DATA_OUTPUT_DIR is now documented in the module setup block and in the README Bright Data environment table, both noting it is optional and defaults to the current working directory.

yonib05:
@strandly-the-agent can you re-review this PR please?

src/strands_tools/bright_data.py
```diff
@@ -88,6 +93,115 @@
...
+        ValueError: If any path component is a symlink.
+    """
+    final_path = str(root.joinpath(*parts))
+    nofollow = getattr(os, "O_NOFOLLOW", 0)
+    # ELOOP is the POSIX result of O_NOFOLLOW hitting a symlink (FreeBSD uses
+    # EMLINK); macOS reports ENOTDIR when O_NOFOLLOW|O_DIRECTORY hits one.
+    symlink_errnos = {errno.ELOOP, errno.ENOTDIR, getattr(errno, "EMLINK", errno.ELOOP)}
+
+    if os.open in os.supports_dir_fd:
+        dir_fd = os.open(str(root), os.O_RDONLY | os.O_DIRECTORY)
```
strandly-the-agent, line 175: 🔴 **The configured root can be replaced before this pathname is reopened.** `root` was resolved at line 131, but after the remote request this call follows whatever now occupies that pathname. I made `requests.post` rename the configured root and replace it with an outside-pointing symlink; twice, the method returned `allowed/shot.png` while the original root stayed empty and `outside/shot.png` contained `b"ATTACK"`.

Please securely acquire and pin the intended root handle before the request, retain it through the response, and create the file relative to that handle while rejecting root-level symlink/reparse substitution. There is no safe small `suggestion` here—the handle lifetime and API between validation/request/write need to change together.

src/strands_tools/bright_data.py
```diff
@@ -88,6 +93,115 @@
...
+            raise
+        finally:
+            os.close(dir_fd)
+    else:
+        probe = root
+        for part in parts:
+            probe = probe / part
+            if probe.is_symlink():
+                raise _confinement_error(root, f"output_path must not contain a symlink: {final_path}.")
+        fd = os.open(
```
strandly-the-agent, line 196: 🔴 **The non-`dir_fd` fallback still has the original check/open race.** Lines 191–195 inspect components, then this separately opens the full pathname. In a forced-fallback repro I replaced `images` with an outside-pointing symlink immediately after its successful `is_symlink()` check; twice, the returned path stayed under `allowed/images`, but `b"ATTACK"` was written outside.

`O_NOFOLLOW` may be unavailable here and, when present, only protects the final component. Because supported Windows uses this branch, please use handle-based/no-reparse traversal there or fail closed and explicitly scope the platform limitation. No small suggestion is safe for this platform-specific handle change.

src/strands_tools/bright_data.py
```diff
@@ -158,25 +272,42 @@ def get_screenshot(self, url: str, output_path: str, zone: Optional[str] = None)
...
+                STRANDS_BRIGHT_DATA_OUTPUT_DIR when set).
             zone (Optional[str]): Override default zone
 
         Returns:
             str: Path to saved screenshot
+
+        Raises:
+            ValueError: If output_path resolves outside the allowed output directory.
         """
+        # Validate the destination before making the (billable) remote request.
```
strandly-the-agent, line 296: 🟡 **Only lexical containment is validated before the billable request.** With an existing final symlink, the current code made one mocked request and only then raised `ValueError` in `_open_confined_output_file()`; I repeated that twice. So this comment and the method docstring’s “destination is validated” claim overstate the behavior.

Prefer securely acquiring/validating the root and destination parent before the request—this also supports the root-handle fix above. Otherwise narrow the wording to say that only lexical containment is checked before requesting.

tests/test_bright_data.py
```diff
@@ -311,3 +311,176 @@ def test_zone_default_fallback():
...
+@patch.dict(os.environ, {"BRIGHTDATA_API_KEY": "test_api_key"})
+def test_get_screenshot_rejects_directory_swapped_for_symlink_after_validation(tmp_path):
+    """A validated directory replaced by a symlink before the write cannot redirect it."""
+    root = tmp_path / "allowed"
+    images = root / "images"
+    images.mkdir(parents=True)
+    outside = tmp_path / "outside"
+    outside.mkdir()
+
+    def swap_then_respond(*args, **kwargs):
```
strandly-the-agent, line 429: 🟡 **This regression passes against the known-vulnerable implementation.** I copied this current test to old head `582f2099`; it passed (`1 passed`) because that version validates only after `requests.post`, sees the already-installed symlink, and never reaches its check/open race.

Synchronize the swap after validation and immediately before the final open/write in the replacement design. The security regression should fail on `582f2099` and pass only when validation and creation are bound safely.

strandly-the-agent (changes requested):
🔴 **Changes requested:** the prior five findings are fixed, but the new secure-open path still has two reproduced confinement escapes.

✅ Re-reviewed exact head `d5286466`; focused tests: 25 passed; `git diff --check` passed.
✅ Latest ordinary exact-head CI is green across Linux/Windows/macOS, dependencies, lint, and CI Gate.
⚠️ Overall rollup failure is the trusted `pull_request_target` checkout guard—not a Bright Data test failure; a superseded run was concurrency-cancelled.
🔴 Root-swap and fallback check/open escapes reproduced twice each.
❓ **API sign-off still needed:** can a Bright Data/codeowner approve the intentional `BrightDataClient.get_screenshot()` behavior change? No approval is visible yet.

<details><summary>Delta review details</summary>

- The previous schema, configured-root error, setup-docs, lexical pre-request rejection, and original child-path findings are resolved; I did not repeat them.
- **Correctness/adversarial:** retained two new confinement bypasses and the remaining billable-request behavior.
- **Test quality:** retained the race regression because it passes against known-vulnerable `582f2099`; suppressed lower-value nested-success/fd-cleanup gaps.
- **LLM context/docs:** generated `output_path` schema is now correct; implementation claims remain too broad only where folded into the security findings.
- Advanced API/adversarial tiers were unavailable due retention validation; both were rerun on the default tier.

No safe small suggestion exists for the handle-lifetime/platform security changes; they need an architectural fix rather than another pathname check.
</details>

PR #524, opened 2026-06-27
fix(code_interpreter): partition session cache by caller key instead of AWS identity

yonib05:
## Description

`AgentCoreCodeInterpreter` keeps a module-level `_session_mapping` cache so that a new instance in a long-running process can reconnect to an existing AWS code interpreter session instead of recreating it. The cache was keyed only on the user-facing `session_name`, which is process-global, so two instances using the same `session_name` would reconnect to the same cached session regardless of who the caller was.

This change partitions the cache by an explicit, caller-supplied `partition_key`. The cache is keyed on `(partition_key, session_name)`, and `partition_key` is bound at construction from operator-controlled input and is never read from a tool action, so a model-supplied `session_name` cannot reach across partitions.

A service typically runs under a single IAM identity while serving many callers, so an AWS-credential-derived key would collapse those callers together. Keying on a caller-supplied partition instead lets one identity host many isolated sessions:

- `partition_key` defaults to a shared `"default"` partition, preserving the prior single-caller reconnect behavior.
- Multi-tenant callers pass a distinct `partition_key` per principal — the authenticated principal/tenant id, or the agent session id (e.g. `context.session_id`) when one identity runs several isolated sessions.

The in-process cache is a reconnect convenience, not the isolation boundary. The real boundary is AWS-side: a session is reachable only by holding its server-issued, unguessable `sessionId`, and every operation is authorized against the interpreter's IAM execution role. Within a single process, `partition_key` is what keeps callers apart.

The public tool contract is unchanged: the `session_name` action field and existing constructor arguments behave as before; `partition_key` is a new optional argument.

## Changes

- Add an optional `partition_key` constructor argument and key `_session_mapping` on `(partition_key, session_name)` via the `_scoped_key` helper.
- Bind `partition_key` at construction (default `"default"`); it is never taken from a tool action.
- Remove the AWS-credential / STS `GetCallerIdentity` derived key (and its eager call at init); the cache key no longer depends on AWS credentials, so credential rotation cannot cause a reconnect miss.
- Reframe the docstring and module comment to state the real boundary (server-issued `sessionId` + IAM execution role) and describe the cache as a per-process reconnect convenience partitioned by `partition_key`.

## Testing

- `pytest tests/code_interpreter/` passes (59 tests).
- `ruff format` / `ruff check` clean.
- Regression tests:
  - default `partition_key` is the shared `"default"` partition,
  - different partitions produce distinct cache keys for the same `session_name`,
  - a session cached in one partition is not reachable from another partition via the same `session_name` (the multi-tenant case),
  - two instances sharing a `partition_key` reconnect to the same cached AWS session,
  - the cache key is independent of the boto session / credentials.

### DIFF
--- a/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
+++ b/src/strands_tools/code_interpreter/agent_core_code_interpreter.py
@@ -21,18 +21,19 @@ from .models import (
 
 logger = logging.getLogger(__name__)
 
-# Module-level session cache - persists across object instances within a single process.
+# Module-level session cache - lets a new object instance in the same process reconnect to
+# an AWS code interpreter session a previous instance started, instead of paying to recreate
+# it. Keys are partitioned per caller (see partition_key / _scoped_key) so that a session name
+# supplied in one caller's partition can never address a session created in another's.
 #
-# Keys are scoped per caller identity (see _resolve_isolation_key) so that instances
-# configured with different AWS credentials do not reconnect to each other's sessions
-# when they happen to use the same session_name. Instances that share the same identity
-# (the common case: one process, one set of credentials) still reconnect as before.
-#
-# NOTE: this in-process cache is a performance optimization, not an isolation boundary.
-# It only persists within a single Python process and is keyed on a best-effort identity.
-# Deployments that serve multiple distinct callers must run them in separate processes and
-# use distinct, unguessable session names. Do not rely on this cache to keep callers apart.
-_session_mapping: Dict[str, str] = {}  # "isolation_key\x00user_session_name" -> aws_session_id
+# The real isolation boundary is AWS-side, not this map: a session is reachable only by holding
+# its server-issued, unguessable sessionId, and every operation is authorized against the
+# interpreter's IAM execution role. This cache is a per-process reconnect convenience. The
+# partition_key is what keeps callers apart within one process; it is bound at construction
+# from operator-controlled input and is never taken from a tool action, so a model-supplied
+# session name cannot reach across partitions. Multi-tenant deployments should pass a distinct
+# partition_key per caller (e.g. the authenticated principal or agent session id).
+_session_mapping: Dict[str, str] = {}  # "partition_key\x00session_name" -> aws_session_id
 
 
 @dataclass
@@ -66,6 +67,7 @@ class AgentCoreCodeInterpreter(CodeInterpreter):
         persist_sessions: bool = True,
         session_timeout_seconds: int = 900,
         boto_session: Optional[boto3.Session] = None,
+        partition_key: Optional[str] = None,
     ) -> None:
         """
         Initialize the Bedrock AgentCore code interpreter with session persistence support.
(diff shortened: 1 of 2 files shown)


src/strands_tools/code_interpreter/agent_core_code_interpreter.py
```diff
@@ -21,8 +21,18 @@
...
-# Module-level session cache - persists across object instances
-_session_mapping: Dict[str, str] = {}  # user_session_name -> aws_session_id
+# Module-level session cache - persists across object instances within a single process.
+#
+# Keys are scoped per caller identity (see _resolve_isolation_key) so that instances
+# configured with different AWS credentials do not reconnect to each other's sessions
+# when they happen to use the same session_name. Instances that share the same identity
+# (the common case: one process, one set of credentials) still reconnect as before.
+#
+# NOTE: this in-process cache is a performance optimization, not an isolation boundary.
```
jsamuel1: If isolation key is for ensuring sessions get their own code interpreter, why do we need this warning?   

What cases will it not keep callers apart?  (eg. "Other tools can read this global, therefore this may be bypassed. Do not treat this as a hard security boundary" )

src/strands_tools/code_interpreter/agent_core_code_interpreter.py
```diff
@@ -208,6 +230,55 @@ def invoke(payload, context):
             f"persist_sessions={persist_sessions}"
         )
 
+    def _resolve_isolation_key(self, boto_session: Optional[boto3.Session]) -> str:
+        """Derive a rotation-stable cache-scoping key for the configured credentials.
```
jsamuel1: This only works as an isolation key between AWS credentials.  Services often run with service credentials and need isolation by the callers Session identifier.   How can this be made generic, such that other partition keys can be used?

tests/code_interpreter/test_agent_core_code_interpreter.py
```diff
@@ -425,8 +425,8 @@ def test_init_session_success(mock_client_class, interpreter, mock_client):
     assert session_info.description == "Test session"
     assert session_info.client == mock_client
 
-    # Check module-level cache
-    assert _session_mapping.get("my-session") == "test-session-id-123"
+    # Check module-level cache (keyed per caller identity)
```
jsamuel1, line 428: Should be keyed per caller session ID where available.  An identity may have multiple sessions that should be isolated.

jsamuel1 (changes requested):

jsamuel1:
Consider the way session_id is used in the SessionManager functionality.

yonib05:
Reworked. The cache is now keyed on `(partition_key, session_name)`, where `partition_key` is an explicit constructor arg bound from operator input and never read from a tool action — so a model-supplied `session_name` can't cross partitions.

Dropped the STS/credential-derived key: one IAM identity serves many callers, so keying on credentials collapsed them. Operators now pass the caller's session id as `partition_key` per request (like `SessionManager` gets `session_id`); defaults to `"default"` for single-caller back-compat.

Re: the warning — the map was never the boundary. Sessions are reachable only via the server-issued unguessable `sessionId`, authorized against the IAM execution role; `partition_key` just separates callers in-process. Docstring updated.

PR #535, opened 2026-07-09
fix(retrieve): support distance score filtering

### LINKED ISSUE #417: [BUG] Retrieve tool incorrectly filters out relevant results when using AWS Aurora PostgreSQL as backing vector database due to cosine distance being used

saime428:
Summary:
- Add a `scoreMetric` option to the retrieve tool so callers can choose similarity or distance score filtering.
- Keep the existing similarity behavior as the default while using `<=` for distance metrics.
- Add regression coverage for distance filtering and invalid score metric values.

Why:
- Some Bedrock Knowledge Base backends return distance scores where lower values are more relevant. The previous `>=` threshold dropped relevant low-distance results and kept worse high-distance results.

Validation:
- [x] `python -m compileall src\strands_tools\retrieve.py`
- [x] `python -m compileall tests\test_retrieve.py`
- [x] `python -m ruff check src\strands_tools\retrieve.py tests\test_retrieve.py`
- [x] `pytest tests\test_retrieve.py -q`

Notes for reviewers:
- Closes #417.
- The default remains `similarity`, so existing callers keep the previous threshold semantics.

### DIFF
--- a/src/strands_tools/retrieve.py
+++ b/src/strands_tools/retrieve.py
@@ -144,12 +144,22 @@ Usage Examples:
                 "score": {
                     "type": "number",
                     "description": (
-                        "Minimum relevance score threshold (0.0-1.0). Results below this score will be filtered out. "
+                        "Score threshold (0.0-1.0). With scoreMetric='similarity', results below this score are "
+                        "filtered out. With scoreMetric='distance', results above this score are filtered out. "
                         "Default is 0.4."
                     ),
                     "default": 0.4,
                     "minimum": 0.0,
                     "maximum": 1.0,
+                },
+                "scoreMetric": {
+                    "type": "string",
+                    "description": (
+                        "How to interpret the score threshold. Use 'similarity' when higher scores are better, "
+                        "or 'distance' when lower scores are better. Default is 'similarity'."
+                    ),
+                    "enum": ["similarity", "distance"],
+                    "default": "similarity",
                 },
                 "profile_name": {
                     "type": "string",
@@ -187,22 +197,37 @@ Usage Examples:
 }
 
 
-def filter_results_by_score(results: List[Dict[str, Any]], min_score: float) -> List[Dict[str, Any]]:
+def filter_results_by_score(
+    results: List[Dict[str, Any]], min_score: float, score_metric: str = "similarity"
+) -> List[Dict[str, Any]]:
     """
-    Filter results based on minimum score threshold.
+    Filter results based on score threshold.
 
     This function takes the raw results from a knowledge base query and removes
-    any items that don't meet the minimum relevance score threshold.
+    any items that don't meet the configured score threshold.
 
     Args:
         results: List of retrieval results from Bedrock Knowledge Base
-        min_score: Minimum score threshold (0.0-1.0). Only results with scores
-            greater than or equal to this value will be returned.
+        min_score: Score threshold (0.0-1.0).
+        score_metric: Score interpretation. ``similarity`` keeps scores greater
+            than or equal to the threshold. ``distance`` keeps scores less than
+            or equal to the threshold.
 
     Returns:
-        List of filtered results that meet or exceed the score threshold
+        List of filtered results that meet the score threshold.
+
+    Raises:
+        ValueError: If score_metric is not supported.
     """
-    return [result for result in results if result.get("score", 0.0) >= min_score]
+    if score_metric == "similarity":
+        return [result for result in results if result.get("score", 0.0) >= min_score]
+    if score_metric == "distance":
+        return [result for result in results if result.get("score", float("inf")) <= min_score]
+    raise ValueError("scoreMetric must be either 'similarity' or 'distance'")
+
+
+def _score_threshold_operator(score_metric: str) -> str:
+    return "<=" if score_metric == "distance" else ">="
 
 
 # Mapping of RetrievalResultLocation types to their document identifier fields.
@@ -219,7 +244,9 @@ _LOCATION_FIELD_MAP = {
 }
 
 
-def format_results_for_display(results: List[Dict[str, Any]], enable_metadata: bool = False) -> str:
+def format_results_for_display(
+    results: List[Dict[str, Any]], enable_metadata: bool = False, score_metric: str = "similarity"
+) -> str:
     """
     Format retrieval results for readable display.
 
@@ -230,12 +257,15 @@ def format_results_for_display(results: List[Dict[str, Any]], enable_metadata: b
     Args:
         results: List of retrieval results from Bedrock Knowledge Base
         enable_metadata: Whether to include metadata in the formatted output (default: False)
+        score_metric: Score interpretation, either "similarity" or "distance" (default: "similarity")
 
     Returns:
         Formatted string containing the results in a readable format, including score,
         document ID, optional metadata, and content.
     """
     if not results:
+        if score_metric == "distance":
+            return "No results found at or below score threshold."
         return "No results found above score threshold."
 
     formatted = []
@@ -296,6 +326,7 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
             knowledgeBaseId: The ID of the knowledge base to query (default: from environment)
             region: AWS region where the knowledge base is located (default: us-west-2)
             score: Minimum relevance score threshold (default: 0.4)
+            scoreMetric: Score interpretation, either "similarity" or "distance" (default: "similarity")
             profile_name: Optional AWS profile name to use
             retrieveFilter: Optional filter to apply to the retrieval results
 
@@ -331,8 +362,16 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
         kb_id = tool_input.get("knowledgeBaseId", default_knowledge_base_id)
         region_name = tool_input.get("region", default_aws_region)
         min_score = tool_input.get("score", default_min_score)
+        score_metric = tool_input.get("scoreMetric", "similarity")
         enable_metadata = tool_input.get("enableMetadata", default_enable_metadata)
         retrieve_filter = tool_input.get("retrieveFilter")
+
+        if score_metric not in {"similarity", "distance"}:
+            return {
+                "toolUseId": tool_use_id,
+                "status": "error",
+                "content": [{"text": "scoreMetric must be either 'similarity' or 'distance'"}],
+            }
 
         # Initialize Bedrock client with optional profile name
         profile_name = tool_input.get("profile_name")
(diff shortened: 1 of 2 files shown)


saime428:
Hi, checking in on this fix for #417. The focused retrieve tests listed in the PR passed locally, the authorization check succeeded, and `check-access-and-checkout` is still waiting. Please let me know if any changes or additional validation would help.

src/strands_tools/retrieve.py
```diff
@@ -331,9 +362,17 @@ def retrieve(tool: ToolUse, **kwargs: Any) -> ToolResult:
         kb_id = tool_input.get("knowledgeBaseId", default_knowledge_base_id)
         region_name = tool_input.get("region", default_aws_region)
         min_score = tool_input.get("score", default_min_score)
+        score_metric = tool_input.get("scoreMetric", "similarity")
```
kimnamu: `MIN_SCORE` is a similarity threshold, but in distance mode it gets reused as a distance *ceiling* — so the env var filters backwards. I ran the branch against mocked pgvector-style scores `[0.05, 0.30, 0.55, 0.90]` (cosine distance, lower = better), with no explicit `score` in the tool input:

```
### distance mode, caller does NOT pass `score` (relies on MIN_SCORE env)
MIN_SCORE=0.4  -> 'Retrieved 2 results with score <= 0.4:'   kept: ['doc-near', 'doc-good']
MIN_SCORE=0.7  -> 'Retrieved 3 results with score <= 0.7:'   kept: ['doc-near', 'doc-good', 'doc-weak']
MIN_SCORE=0.9  -> 'Retrieved 4 results with score <= 0.9:'   kept: ['doc-near', 'doc-good', 'doc-weak', 'doc-junk']
```

Raising `MIN_SCORE` is "stricter" for every existing similarity user, but here it *loosens* the filter until the near-orthogonal 0.90 chunk comes back. So a deployment that already sets `MIN_SCORE=0.9` and flips to `scoreMetric="distance"` gets the same garbage-in-context symptom #417 reports, just from the other direction.

This is the one thing that makes me want the default decoupled from the metric:

```suggestion
        score_metric = tool_input.get("scoreMetric", "similarity")
        # MIN_SCORE is a similarity threshold; as a distance ceiling it would filter backwards,
        # so distance mode keeps everything unless the caller passes an explicit score.
        default_score = default_min_score if score_metric == "similarity" else float("inf")
        min_score = tool_input.get("score", default_score)
```

Verified with that applied: `ruff format --check` (1 file already formatted), `ruff check` (All checks passed!), `pytest tests/test_retrieve.py` (32 passed). Behaviour after the patch — the env var no longer leaks in:

```
MIN_SCORE=0.4  -> 'Retrieved 4 results with score <= inf:'
MIN_SCORE=0.7  -> 'Retrieved 4 results with score <= inf:'
MIN_SCORE=0.9  -> 'Retrieved 4 results with score <= inf:'
```

Similarity mode is byte-identical before and after (`>= 0.4` keeps `['doc-weak','doc-junk']`, `>= 0.9` keeps `['doc-junk']` in both). #443 landed on the same `float("inf")` default independently, which is some evidence it's the natural choice.

The cost of `inf` is that distance users lose a way to set a default ceiling from the environment — if you'd rather keep one, a separate `RETRIEVE_DISTANCE_MAX_SCORE` (or documenting that `MIN_SCORE` is read as a ceiling in distance mode) both work. What I'd avoid is the current state, where one env var silently means "floor" or "ceiling" depending on another parameter.

A regression test in the existing `mock.patch.dict(os.environ, ...)` style — this one asserts the *invariant* (tightening `MIN_SCORE` must never widen a distance result set) rather than a specific count, so it can't pass in both directions:

```python
def test_retrieve_distance_default_threshold_ignores_similarity_min_score(mock_boto3_client):
    """Tightening the similarity-tuned MIN_SCORE must never widen a distance-mode result set."""
    tool_use = {
        "toolUseId": "test-tool-use-id",
        "input": {
            "text": "test query",
            "knowledgeBaseId": "test-kb-id",
            "scoreMetric": "distance",
            # No explicit score - falls back to the default threshold
        },
    }

    counts = []
    for min_score in ("0.4", "0.7", "0.9"):
        with mock.patch.dict(os.environ, {"MIN_SCORE": min_score}):
            result = retrieve.retrieve(tool=tool_use)
        assert result["status"] == "success"
        text = result["content"][0]["text"]
        counts.append(sum(doc in text for doc in ("doc-001", "doc-002", "doc-003")))

    assert counts[0] == counts[1] == counts[2], f"MIN_SCORE leaked into the distance ceiling: {counts}"
```

Red/green on your branch — without the fix:

```
E       AssertionError: MIN_SCORE leaked into the distance ceiling: [1, 2, 3]
E       assert 1 == 2
tests/test_retrieve.py:819: AssertionError
1 failed, 32 deselected
```

and with it: `33 passed`.

kimnamu (commented):
Thanks for picking this up, @saime428 — the `score_metric` split is the right shape for #417, and keeping `"similarity"` as the default means nobody's existing KB setup shifts under them. I'm not a maintainer, just another Bedrock Knowledge Base user; I checked out `109b179` and ran everything below locally.

**What I confirmed holds**

- The inversion #417 describes is real, and this fixes it. With pgvector `<=>` cosine-distance scores `[0.05, 0.30, 0.55, 0.90]` and threshold `0.4`, the old `>=` path keeps `{0.55, 0.90}` — it drops the near-exact 0.05 match and keeps the near-orthogonal one. Your `distance` path keeps `{0.05, 0.30}`.
- No regression for similarity users. Same fixture, same thresholds: `>= 0.4` → `['doc-weak','doc-junk']`, `>= 0.7` → `['doc-junk']`, `>= 0.9` → `['doc-junk']`, identical to `main`. The 29 pre-existing tests still pass.
- Gates are green as submitted: `ruff format --check` (2 files already formatted), `ruff check` (All checks passed!), `pytest tests/test_retrieve.py` → `32 passed`.
- Your three new tests actually catch the bug. I kept the tests, reverted only `src/strands_tools/retrieve.py` to `HEAD~1`, and got `3 failed, 29 passed`:

```
E   TypeError: filter_results_by_score() got an unexpected keyword argument 'score_metric'
E   AssertionError: assert 'Retrieved 1 results with score <= 0.4' in 'Retrieved 2 results with score >= 0.4:\n\nScore: 0.9000\nDocument ID: doc-001\n...'
E   AssertionError: assert 'success' == 'error'
```

Restoring the source → `32 passed`. So the red/green story is solid.

**One thing I'd change before merge** (inline suggestion on `retrieve.py:365`): in `distance` mode the default threshold still comes from `MIN_SCORE`, which is a *similarity* knob. Raising it loosens the distance filter instead of tightening it, so an existing `MIN_SCORE=0.9` deployment that flips to `scoreMetric="distance"` gets the same bad-chunks-in-context symptom #417 reports. Repro, patch, `ruff`/`pytest` output, and a red/green regression test are in the inline comment. #443 arrived at the same `float("inf")` default independently.

**Two non-blocking notes**

1. `score` keeps `"maximum": 1.0`, but cosine distance ranges `[0, 2]` — [pgvector's README](https://github.com/pgvector/pgvector#vector-operators) documents `<=>` as cosine distance, and cosine similarity is derived as `1 - distance`, so 2.0 is reachable for opposing vectors. #417 flagged this too ("the `score` parameter schema constrains to `[0.0, 1.0]`, but cosine distance ranges `[0, 2]`"). It isn't enforced at runtime — I passed `score=1.5` both directly and through `agent.tool.retrieve(...)` and both returned `Retrieved 2 results with score <= 1.5:` with the 1.35 chunk included — so it's a docs/schema-honesty issue rather than a functional block. Bumping `maximum` to `2.0` (or dropping it in favour of prose) would stop the schema from telling the model something the tool doesn't actually believe.
2. `scoreMetric` is camelCase, matching `numberOfResults`/`knowledgeBaseId`/`enableMetadata` in this spec, so it's the consistent choice — worth noting only because #443 uses `score_metric`, and whichever lands second will want to match.

**On the design debate in #443** — @starJammer's concern there ("don't make the LLM reason about similarity vs distance") applies to this PR's `inputSchema` too, since `scoreMetric` is model-visible either way. My read is that the filtering core here is correct regardless of how the metric gets selected, and the selection surface is a separate, smaller decision: an env var read alongside `MIN_SCORE` (developer-controlled, invisible to the model) would reuse this exact code. Entirely a maintainer call, and I don't think it should hold up the correctness fix. Worth deciding which of #535 / #443 is the base, though — they overlap almost completely.

The scope is tight and the tests are real. With the default-threshold tweak I think this is ready. Thanks again for the fix.

src/strands_tools/retrieve.py
saime428: Good catch. I changed distance mode to use `inf` when `score` is omitted, so `MIN_SCORE` now only controls the similarity default. I also added regression coverage for `MIN_SCORE` values 0.4, 0.7, and 0.9. `ruff format --check`, `ruff check`, and `pytest tests/test_retrieve.py` all pass (35 tests). Updated in 8f9948b. Thanks for the detailed repro.

lizradway:
Hi! Could you add support for this onto our [memory manager](https://strandsagents.com/docs/user-guide/concepts/memory/overview/)? 

We have shifted our support for memory to that construct. Thanks!

saime428:
Thanks — happy to move this over. I’m planning to put the filtering in `BedrockKnowledgeBaseStore.search()` as developer-configured store options (`score_threshold` and `score_metric`). That would cover both `search_memory` and automatic memory injection without asking the model to choose the metric, and filtering would stay off by default to preserve the current behavior.

Does that direction match what you had in mind? Also, would you prefer Python-only first, or Python and TypeScript support in the same PR?

PR #531, opened 2026-07-06
fix(environment): add STRANDS_DISABLE_LOAD_TOOL to PROTECTED_VARS

gautamsirdeshmukh:
## Description

Adds `STRANDS_DISABLE_LOAD_TOOL` to the environment tool's `PROTECTED_VARS` set, preventing agents from overwriting or deleting the hardening flag via the environment tool. 

Also adds appropriate unit tests.

## Type of Change

Bug fix

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [x] I ran `hatch run prepare`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #406, opened 2026-02-25
fix: add BYPASS_TOOL_CONSENT to PROTECTED_VARS

lizradway:
## Description

`BYPASS_TOOL_CONSENT` globally disables user consent prompts across the framework but wasn't in the environment tool's `PROTECTED_VARS` blocklist. This allowed an agent to set it via the environment tool with a single user approval, silently disabling all consent gates — including the environment tool's own. This fix adds it to `PROTECTED_VARS` so it can't be modified at runtime.

## Related Issues

n/a

## Documentation PR

n/a

## Type of Change

Bug fix

## Testing

How have you tested the change?  Verify that the changes do not break functionality or introduce warnings in consuming repositories: agents-docs, agents-tools, agents-cli

- [X] I ran `hatch run prepare`

## Checklist
- [X] I have read the CONTRIBUTING document
- [X] I have added any necessary tests that prove my fix is effective or my feature works
- [X] I have updated the documentation accordingly
- [X] I have added an appropriate example to the documentation to outline the feature
- [X] My changes generate no new warnings
- [X] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #545, opened 2026-07-27
fix(environment): drive value masking from environment variable

yonib05:
## Description

The `environment` tool declared `masked` on its `inputSchema` and read it from `tool_input` on the `list` and `get` actions. Because `inputSchema` is the interface the agent uses to invoke the tool, the agent could pass `masked=false` and have sensitive values returned in cleartext, overriding the `ENV_VARS_MASKED_DEFAULT` setting the operator configured.

Masking is now read only from `ENV_VARS_MASKED_DEFAULT`, matching how `python_repl` sources its non-interactive mode in #541.

Changes:

- Removed the `masked` property from `TOOL_SPEC["inputSchema"]`.
- Resolve `masked` once from `ENV_VARS_MASKED_DEFAULT` (default `true`) and use it for both `list` and `get`; removed the two `tool_input.get("masked", ...)` reads.
- Dropped the redundant `final_display_value` in `get`, which recomputed the already-masked value.
- Documented `ENV_VARS_MASKED_DEFAULT` in the module docstring and clarified the README entry.

`ENV_VARS_MASKED_DEFAULT` keeps its existing name, default, and behavior, so operators who set it see no change. Callers that passed `masked` no longer have any effect from it.

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

`hatch run prepare` passes. The one failure I see locally, `tests/test_agent_core_memory.py::test_initialization`, also fails on an unmodified `main` in my environment because it picks up my `AWS_REGION`, and is unrelated to this change.

Updated `test_environment_masked_values` to disable masking through the environment variable rather than tool input, and added three tests:

- `test_masked_not_in_input_schema` asserts `masked` is absent from the input schema.
- `test_get_ignores_masked_tool_input` and `test_list_ignores_masked_tool_input` pass `masked=False` in tool input and assert the sensitive value is still masked.

## Checklist

----

### DIFF
--- a/src/strands_tools/environment.py
+++ b/src/strands_tools/environment.py
@@ -61,7 +61,9 @@ agent.tool.environment(action="delete", name="TEMP_VAR")
 Configuration:
 - ENV_VARS_MASKED_DEFAULT (environment variable): Set to "false" to return sensitive
   values unmasked. Masking is configured by the operator only; it cannot be changed
-  through tool input, and the variable is protected so the tool cannot set it either.
+  through tool input, and this tool cannot set it. Note that tools which execute
+  arbitrary code in the agent process, such as python_repl and shell, can still write
+  the process environment, so isolate those separately.
 
 See the environment function docstring for more details on available actions and parameters.
 """
@@ -160,7 +162,18 @@ PROTECTED_VARS = {
     "STRANDS_NON_INTERACTIVE",
     "STRANDS_DISABLE_LOAD_TOOL",
     "ENV_VARS_MASKED_DEFAULT",
+    "STRANDS_HTTP_ALLOW_INSECURE_SSL",
+    "EDITOR_DISABLE_BACKUP",
 }
+
+
+def is_protected(name: str) -> bool:
+    """Check whether a variable name is protected from modification.
+
+    Windows upper-cases environment variable names on write, so a case-sensitive
+    check would let "path" through and store it as "PATH".
+    """
+    return (name.upper() if os.name == "nt" else name) in PROTECTED_VARS
 
 
 def mask_sensitive_value(name: str, value: str) -> str:
@@ -208,7 +221,7 @@ def format_env_vars_table(env_vars: Dict[str, str], masked: bool, prefix: Option
         if prefix and not name.startswith(prefix):
             continue
 
-        protected = "🔒" if name in PROTECTED_VARS else ""
+        protected = "🔒" if is_protected(name) else ""
         display_value = mask_sensitive_value(name, value) if masked else value
         table.add_row(protected, name, str(display_value))
 
@@ -252,7 +265,7 @@ def format_operation_preview(
     table.add_row("Action", f"[{action_style}]{action.upper()}[/{action_style}]")
 
     if name:
-        protected = name in PROTECTED_VARS
+        protected = is_protected(name)
         name_style = "red" if protected else "white"
         table.add_row(
             "Variable",
@@ -264,7 +277,7 @@ def format_operation_preview(
         table.add_row("Prefix Filter", prefix)
 
     # Add warning for protected variables
-    if name and name in PROTECTED_VARS:
+    if name and is_protected(name):
         table.add_row(
             "⚠️ Warning",
             "[red]This is a protected system variable that cannot be modified[/red]",
@@ -317,7 +330,7 @@ def format_env_vars(env_vars: Dict[str, str], masked: bool, prefix: Optional[str
             {
                 "name": name,
                 "value": mask_sensitive_value(name, value) if masked else value,
-                "protected": name in PROTECTED_VARS,
+                "protected": is_protected(name),
             }
         )
 
@@ -426,7 +439,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
 
     Notes:
         - The ENV var "BYPASS_TOOL_CONSENT" can be set to "true" to bypass confirmation prompts
-        - Protected variables include PATH, PYTHONPATH, STRANDS_HOME, SHELL, USER, HOME
+        - Protected variables are listed in PROTECTED_VARS and cannot be set or deleted
         - Sensitive variables are detected by keywords in their names (TOKEN, SECRET, etc.)
         - Masking is controlled by the ENV var "ENV_VARS_MASKED_DEFAULT" (default: "true")
     """
@@ -444,7 +457,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     tool_input = tool["input"]
 
     # Masking is operator-controlled and cannot be overridden through tool input
-    masked = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").strip().lower() != "false"
+    masked = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").lower() != "false"
 
     # Check for BYPASS_TOOL_CONSENT mode
     strands_dev = os.environ.get("BYPASS_TOOL_CONSENT", "").lower() == "true"
@@ -523,7 +536,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
 
             # Add variable details
             table.add_row("Name", name)
-            table.add_row("Type", "Protected" if name in PROTECTED_VARS else "Standard")
+            table.add_row("Type", "Protected" if is_protected(name) else "Standard")
             table.add_row("Value", display_value)
 
             # Add value properties
@@ -537,10 +550,10 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
             panel = Panel(
                 table,
                 title=(
-                    f"[bold {'yellow' if name in PROTECTED_VARS else 'blue'}]🔍 "
-                    f"Environment Variable Details[/bold {'yellow' if name in PROTECTED_VARS else 'blue'}]"
+                    f"[bold {'yellow' if is_protected(name) else 'blue'}]🔍 "
+                    f"Environment Variable Details[/bold {'yellow' if is_protected(name) else 'blue'}]"
                 ),
-                border_style="yellow" if name in PROTECTED_VARS else "blue",
+                border_style="yellow" if is_protected(name) else "blue",
                 box=box.ROUNDED,
             )
             console.print(panel)
@@ -565,7 +578,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
             value = tool_input["value"]
 
             # Check protected status first, regardless of confirmation mode
-            if name in PROTECTED_VARS:
+            if is_protected(name):
                 error_msg = f"⚠️ Cannot modify protected variable: {name}"
                 error_details = "\nProtected variables ensure system stability and security."
                 console.print(format_error_message(f"{error_msg}{error_details}"))
@@ -674,7 +687,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
             name = tool_input["name"]
 
             # Check protected status first
-            if name in PROTECTED_VARS:
+            if is_protected(name):
                 error_msg = (
                     f"⚠️ Cannot delete protected variable: {name}\n"
                     "Protected variables ensure system stability and security."
--- a/src/strands_tools/environment.py
+++ b/src/strands_tools/environment.py
@@ -58,6 +58,11 @@ agent.tool.environment(action="set", name="MY_SETTING", value="new_value")
 agent.tool.environment(action="delete", name="TEMP_VAR")
 ```
 
+Configuration:
+- ENV_VARS_MASKED_DEFAULT (environment variable): Set to "false" to return sensitive
+  values unmasked. Masking is configured by the operator only; it cannot be changed
+  through tool input.
+
 See the environment function docstring for more details on available actions and parameters.
 """
 
@@ -133,11 +138,6 @@ Key Features:
                 "prefix": {
                     "type": "string",
                     "description": "Filter variables by prefix",
-                },
-                "masked": {
-                    "type": "boolean",
-                    "description": "Mask sensitive values in output",
-                    "default": True,
                 },
             },
             "required": ["action"],
@@ -400,7 +400,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     Security Features:
     ---------------
     - Protected system variables cannot be modified
-    - Sensitive values are masked in output by default
+    - Sensitive values are masked in output unless ENV_VARS_MASKED_DEFAULT is "false"
     - Destructive actions require explicit confirmation
     - Clear risk level indicators for all operations
     - BYPASS_TOOL_CONSENT mode controls for testing and automation
@@ -411,7 +411,6 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
             tool["input"]["name"]: Environment variable name (for get/set/delete/validate)
             tool["input"]["value"]: Value to set (for set action)
             tool["input"]["prefix"]: Filter prefix for list action
-            tool["input"]["masked"]: Whether to mask sensitive values (default: True)
         **kwargs: Additional keyword arguments (unused)
 
     Returns:
@@ -424,7 +423,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
         - The ENV var "BYPASS_TOOL_CONSENT" can be set to "true" to bypass confirmation prompts
         - Protected variables include PATH, PYTHONPATH, STRANDS_HOME, SHELL, USER, HOME
         - Sensitive variables are detected by keywords in their names (TOKEN, SECRET, etc.)
-        - For security reasons, values of sensitive variables are masked in output
+        - Masking is controlled by the ENV var "ENV_VARS_MASKED_DEFAULT" (default: "true")
     """
     console = console_util.create()
 
@@ -439,8 +438,8 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     tool_use_id = tool["toolUseId"]
     tool_input = tool["input"]
 
-    # Get environment variables at runtime
-    env_vars_masked_default = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").lower() == "true"
+    # Masking is operator-controlled and cannot be overridden through tool input
+    masked = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").lower() == "true"
 
     # Check for BYPASS_TOOL_CONSENT mode
     strands_dev = os.environ.get("BYPASS_TOOL_CONSENT", "").lower() == "true"
@@ -460,7 +459,6 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
 
         if action == "list":
             prefix = tool_input.get("prefix")
-            masked = tool_input.get("masked", env_vars_masked_default)
 
             # Format rich table
             table = format_env_vars_table(dict(os.environ), masked=masked, prefix=prefix)
@@ -507,7 +505,6 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
                     "content": [{"text": error_msg}],
                 }
 
-            masked = tool_input.get("masked", env_vars_masked_default)
             safe_value = value if value is not None else ""
             display_value = mask_sensitive_value(name, safe_value) if masked else safe_value
 
@@ -546,8 +543,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
             # Show success message
             show_operation_result(console, True, f"Successfully retrieved {name}")
             # Create a return object with properly cast types
-            final_display_value = display_value if masked else safe_value
-            get_content: List[ToolResultContent] = [{"text": f"{name} = {final_display_value}"}]
+            get_content: List[ToolResultContent] = [{"text": f"{name} = {display_value}"}]
             return {
                 "toolUseId": tool_use_id,
                 "status": "success",
--- a/src/strands_tools/environment.py
+++ b/src/strands_tools/environment.py
@@ -61,7 +61,7 @@ agent.tool.environment(action="delete", name="TEMP_VAR")
 Configuration:
 - ENV_VARS_MASKED_DEFAULT (environment variable): Set to "false" to return sensitive
   values unmasked. Masking is configured by the operator only; it cannot be changed
-  through tool input.
+  through tool input, and the variable is protected so the tool cannot set it either.
 
 See the environment function docstring for more details on available actions and parameters.
 """
@@ -101,8 +101,10 @@ Key Features:
    - Protected variables list
    - Value validation
    - Change tracking
-   - Variable masking
-   
+   - Variable masking: values whose name matches TOKEN/SECRET/PASSWORD/KEY/AUTH are
+     masked in get/list output (e.g. "abcd...5678") as an operator-controlled policy.
+     There is no tool input to disable this.
+
 4. Usage Examples:
    # List all environment variables:
    environment(action="list")
@@ -157,6 +159,7 @@ PROTECTED_VARS = {
     "BYPASS_TOOL_CONSENT",
     "STRANDS_NON_INTERACTIVE",
     "STRANDS_DISABLE_LOAD_TOOL",
+    "ENV_VARS_MASKED_DEFAULT",
 }
 
 
@@ -385,7 +388,8 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     1. The function processes the requested action (list, get, set, delete, validate)
     2. For destructive actions, it requires user confirmation unless in BYPASS_TOOL_CONSENT mode
     3. Protected system variables are identified and cannot be modified
-    4. Sensitive values (tokens, passwords, etc.) are automatically masked
+    4. Sensitive values (tokens, passwords, etc.) are masked by default, controlled by
+       ENV_VARS_MASKED_DEFAULT
     5. Rich output formatting provides clear visual feedback on operations
     6. All operations return structured results for both human and programmatic use
 
@@ -400,7 +404,8 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     Security Features:
     ---------------
     - Protected system variables cannot be modified
-    - Sensitive values are masked in output unless ENV_VARS_MASKED_DEFAULT is "false"
+    - Values of variables whose name contains TOKEN, SECRET, PASSWORD, KEY, or AUTH are
+      masked in output unless ENV_VARS_MASKED_DEFAULT is set to "false"
     - Destructive actions require explicit confirmation
     - Clear risk level indicators for all operations
     - BYPASS_TOOL_CONSENT mode controls for testing and automation
@@ -439,7 +444,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     tool_input = tool["input"]
 
     # Masking is operator-controlled and cannot be overridden through tool input
-    masked = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").lower() == "true"
+    masked = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").strip().lower() != "false"
 
     # Check for BYPASS_TOOL_CONSENT mode
     strands_dev = os.environ.get("BYPASS_TOOL_CONSENT", "").lower() == "true"
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -246,11 +246,39 @@ def test_environment_masked_values(agent, os_environment):
     # The full value should not appear in the output
     assert sensitive_value not in extract_result_text(result)
 
-    # Test with masking disabled
-    result = agent.tool.environment(action="get", name=sensitive_name, masked=False)
+    # Test with masking disabled by the operator
+    os_environment["ENV_VARS_MASKED_DEFAULT"] = "false"
+    result = agent.tool.environment(action="get", name=sensitive_name)
     assert result["status"] == "success"
     # Now the full value should appear
     assert sensitive_value in extract_result_text(result)
+
+
+def test_masked_not_in_input_schema():
+    """Masking must not be configurable through the model-facing tool input schema."""
+    assert "masked" not in environment.TOOL_SPEC["inputSchema"]["json"]["properties"]
+
+
+def test_get_ignores_masked_tool_input(agent, os_environment):
+    """A masked value in tool input must not unmask a sensitive value."""
+    sensitive_name = "TEST_TOKEN_SECRET"
+    sensitive_value = "abcd1234efgh5678"
+    os_environment[sensitive_name] = sensitive_value
+
+    result = agent.tool.environment(action="get", name=sensitive_name, masked=False)
+    assert result["status"] == "success"
+    assert sensitive_value not in extract_result_text(result)
+
+
+def test_list_ignores_masked_tool_input(agent, os_environment):
+    """A masked value in tool input must not unmask sensitive values in a listing."""
+    sensitive_name = "TEST_TOKEN_SECRET"
+    sensitive_value = "abcd1234efgh5678"
+    os_environment[sensitive_name] = sensitive_value
+
+    result = agent.tool.environment(action="list", masked=False)
+    assert result["status"] == "success"
+    assert sensitive_value not in extract_result_text(result)
 
 
 def test_direct_set_protected_var_strands_disable_load_tool(agent, os_environment):
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -301,9 +301,9 @@ def test_cannot_set_masking_variable(agent, os_environment):
     assert sensitive_value not in extract_result_text(result)
 
 
-@pytest.mark.parametrize("value", ["1", "yes", "on", " true", "true ", ""])
+@pytest.mark.parametrize("value", ["1", "yes", "on", " true", "true ", "", " false", "0", "no", "off"])
 def test_masking_stays_on_for_non_false_values(agent, os_environment, value):
-    """Only an explicit "false" disables masking, so a typo cannot unmask values."""
+    """Only an exact "false" disables masking, so a typo or stray space cannot unmask values."""
     sensitive_name = "TEST_TOKEN_SECRET"
     sensitive_value = "abcd1234efgh5678"
     os_environment[sensitive_name] = sensitive_value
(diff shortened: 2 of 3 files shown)


yonib05:
@strandly-the-agent can you review this PR?

strandly-the-agent (commented):
**Changes requested — one blocker: the guarantee this PR adds is falsifiable in one extra tool call.**

The core fix is right, and I verified it: `masked=false` from the model is now ignored on both `get` and `list` (on `main` the same call returned cleartext). But `ENV_VARS_MASKED_DEFAULT` isn't in `PROTECTED_VARS`, so the agent can `set` it to `"false"` and read secrets in cleartext on the next call — which makes the new docstring line *"it cannot be changed through tool input"* and the README's *"the only way to configure masking"* untrue. Two one-line changes close that and the fail-open parse; I ran them and the suite stays green.

**Verified on `83c4dff`** (sandbox, `hatch` envs):

| | |
|---|---|
| ✅ | `hatch test tests/test_environment.py` → **22 passed**; `hatch run lint` → All checks passed |
| ✅ | the fix holds — model-supplied `masked=False` ignored on `get` *and* `list` |
| 🔴 | self-set bypass reproduced → `set ENV_VARS_MASKED_DEFAULT=false` then `get` → `MY_TOKEN = SUPER_SECRET_VALUE_9876` |
| ✅ | proposed 2-line fix → `set` returns `Cannot modify protected variable`, `get` stays `SUPE...9876`, **22 still pass** |

4 inline comments: 1 🔴 blocker, 2 🟡 should-fix, 1 ⚪ nit. Posting as a comment-review rather than a formal Request-Changes — I'm an experimental agent, the call is yours.

<details>
<summary><b>Blocker repro (verbatim)</b></summary>

```
$ ENV_VARS_MASKED_DEFAULT unset, BYPASS_TOOL_CONSENT=true, on 83c4dff
ENV_VARS_MASKED_DEFAULT in PROTECTED_VARS? False
masked in inputSchema?                     False
1) get, model tries masked=False : 'MY_TOKEN = SUPE...9876'          <- the fix works
2) set ENV_VARS_MASKED_DEFAULT   : success 'Set ENV_VARS_MASKED_DEFAULT = false'
3) get again                     : 'MY_TOKEN = SUPER_SECRET_VALUE_9876'
   CLEARTEXT IN MODEL CONTEXT: True
```

Unconditional under `BYPASS_TOOL_CONSENT=true` (the automation posture documented at `environment.py:423`). In interactive mode it needs one approval whose panel reads `Set ENV_VARS_MASKED_DEFAULT = false` — nothing tells the approver that authorises secret exfiltration on the *next* call. Not a regression (on `main` the model had the direct `masked=false` path anyway) — it blocks because the PR *asserts* the hole is closed in two operator-facing places.

After adding `"ENV_VARS_MASKED_DEFAULT"` to `PROTECTED_VARS` and using the fail-closed parse:

```
set -> error 'Cannot modify protected variable: ENV_VARS_MASKED_DEFAULT'
get -> 'MY_TOKEN = SUPE...9876' | LEAKED: False
22 passed in 3.46s
```
</details>

<details>
<summary><b>Questions (all non-blocking)</b></summary>

- **Retitle for the release notes?** Squash-merge makes the title the release-notes bullet verbatim (`.github/scripts/group-release-notes.py`), and "drive value masking from environment variable" doesn't signal that a tool-input parameter went away. #513 titled itself *"remove proxies from LLM-controllable tool input schema"* — worth mirroring? (Also: #541, cited as the precedent, is weaker than it looks — `non_interactive_mode` was never in `python_repl`'s schema, so nothing public broke there. #513 is the real structural twin, and this PR matches it.)
- **`ENV_VARS_MASKED_DEFAULT` name** — post-PR it's the only `*_DEFAULT` env var in the repo with nothing left to default *over* (`FILE_READ_RECURSIVE_DEFAULT`, `RETRIEVE_ENABLE_METADATA_DEFAULT`, `SPEAK_DEFAULT_PLAY_AUDIO` all still pair with a live tool-input override). Mild smell, but I'd keep the name — the compat promise beats the churn.
- **Worth a line in the description** on why `environment` keeps an operator escape hatch while #520 made `use_aws` redaction unconditional? The distinction reads right to me (deliberate `get` of a named var vs. secrets riding along in a response nobody asked for) — just currently implicit between two PRs three days apart.
- No `needs-api-review` gate applies: I checked the label set and this repo has no API-review labels. The `Type of Change` box (`Bug fix`) is the mechanism here, and that's your call.
</details>

<details>
<summary><b>Out of scope — pre-existing, suggest one follow-up issue (not this PR)</b></summary>

Found while testing, all present on `main` and untouched by this diff. Flagging so they aren't mistaken for something #545 introduced:

1. **The protected-var tests are vacuous.** Emptying `PROTECTED_VARS` to `set()` leaves all 22 tests passing — `tests/test_environment.py:82-90` and `:126-135` pass on the autouse deny-confirmation fixture (`"n"`), not on protection. With consent mocked to `"y"` and the same mutation, `PATH` is actually overwritten to `/bad/path`. Directly relevant here: the suite **cannot verify the blocker's fix** either way.
2. **`test_environment_dev_mode_protected_var` (`:221-233`) passes on a crash.** It sets `os_environment["BYPASS_TOOL_CONSENT"] = True` (a Python bool), so `environment.py:445` `.lower()` raises `AttributeError: 'bool' object has no attribute 'lower'`, caught by the blanket handler — `assert result["status"] == "error"` is satisfied by the exception, never reaching the protected-var branch. Its second assertion compares a dict to a string, so it's trivially true. Fix: `"true"` as a string, and `assert os_environment["PATH"] == unchanging_value`.
3. **`mask_sensitive_value` has no direct unit test.** Mutating the format to `value[:8] + "..." + value[-8:]` renders the 16-char fixture as `abcd1234...efgh5678` — every character of the secret — and all 22 tests still pass, because the assertions only check the raw string is absent.
4. **The name heuristic misses most real secret names.** With masking fully ON, `DATABASE_URL`, `MONGODB_URI`, `REDIS_URL`, `GH_PAT`, `DB_DSN` and friends come back in cleartext — 10 of 12 realistic names in one `list` call. Worth either extending the pattern or scoping the docs' promise (see inline on `:403`).

Happy to file that as a single issue if useful.
</details>

<details>
<summary><b>What I checked that came back clean</b></summary>

- **Dropping `final_display_value` is a true no-op.** `final_display_value = display_value if masked else safe_value` was a tautology — when `masked` is `False`, `display_value` *already is* `safe_value`. Confirmed over an exhaustive truth table (`masked` × `{None, "", short, 32-char}`): 8/8 identical.
- **The single resolution point reaches every consumer**: `:464` table, `:477` plain text, `:509` `get` (preview, details table, returned content). No stale `env_vars_masked_default`, no leftover `tool_input.get("masked")`, no unused-name lint.
- **No other unmask path into the model's context**, with masking on: `get`, `list` (table *and* plain text), `validate`, `set`, `delete` all keep the raw value out of the returned `ToolResult`. `set`'s "Current vs New" table and `delete`'s "Previous Value" row are console-only (and `console_util.create()` discards output unless `STRANDS_TOOL_CONSOLE_MODE=enabled`).
- **No laundering**: `$MY_TOKEN`, `${MY_TOKEN}`, `%MY_TOKEN%`, `` `echo $MY_TOKEN` ``, `$(...)` are stored literally — the model can't copy a secret into a non-sensitive name.
- **No concurrency issue**: 400 `list` calls racing a flag-flipping thread → 0 partially-masked results. Resolving `masked` once at `:442` is the safer choice.
- **`BYPASS_TOOL_CONSENT` *is* protected**, so the model can't self-grant the bypass; and `delete`ing the masking var restores the safe default.
- **The new tests bite**: reverting the `get` override, the `list` override, or re-adding the schema property each fails exactly one targeted test. `masked` is gone from the model-visible `inputSchema` (`['action','name','prefix','value']`), and no stale `masked` tool-input references remain anywhere in the repo.

Method: staged pipeline — triage/route → context build → 6 parallel fresh-context passes (correctness, API/DevX, adversarial, test-quality, LLM-context, docs) → independent aggregation, which cut ~20 raw findings to the 4 posted. I re-ran the blocker, the proposed fix, the parse matrix, the model-visible spec dump, tests and lint myself before posting.
</details>

src/strands_tools/environment.py
```diff
@@ -58,6 +58,11 @@
 agent.tool.environment(action="delete", name="TEMP_VAR")
 ```
 
+Configuration:
+- ENV_VARS_MASKED_DEFAULT (environment variable): Set to "false" to return sensitive
+  values unmasked. Masking is configured by the operator only; it cannot be changed
+  through tool input.
```
strandly-the-agent: 🔴 **This guarantee doesn't hold — the agent can flip the switch itself.**

`ENV_VARS_MASKED_DEFAULT` isn't in `PROTECTED_VARS` (`environment.py:150-159`), and masking is re-read from `os.getenv` on every call (`:442`), so `set` on this very tool is a second, still-open path to the outcome the PR removes. Verified on `83c4dff`:

```
ENV_VARS_MASKED_DEFAULT in PROTECTED_VARS? False
1) get, model tries masked=False : 'MY_TOKEN = SUPE...9876'          <- your fix works
2) set ENV_VARS_MASKED_DEFAULT   : success 'Set ENV_VARS_MASKED_DEFAULT = false'
3) get again                     : 'MY_TOKEN = SUPER_SECRET_VALUE_9876'
   CLEARTEXT IN MODEL CONTEXT: True
```

Unconditional under `BYPASS_TOOL_CONSENT=true`. In interactive mode it costs one approval that reads `Set ENV_VARS_MASKED_DEFAULT = false` — which doesn't tell the human they're authorising an unmasked read on the *next* call. Not a regression versus `main`, but it does falsify this sentence and the README's "only way to configure masking".

The fix is one line in `PROTECTED_VARS`, same shape as #531 for `STRANDS_DISABLE_LOAD_TOOL` and #406 for `BYPASS_TOOL_CONSENT`:

```python
 PROTECTED_VARS = {
     ...
     "STRANDS_DISABLE_LOAD_TOOL",
+    "ENV_VARS_MASKED_DEFAULT",
 }
```

I applied it: `set` → `error 'Cannot modify protected variable: ENV_VARS_MASKED_DEFAULT'`, `get` → `'MY_TOKEN = SUPE...9876'`, and `hatch test tests/test_environment.py` → 22 passed.

Worth a regression test alongside it (`set` on the var errors, and a following `get` stays masked) — I mutated `PROTECTED_VARS` to `set()` and all 22 tests still passed, so nothing currently guards this.

If you'd rather keep the diff strictly to the schema change, the alternative is softening this sentence and `README.md:1220` to say the var isn't protected — but that trades a one-line fix for a weaker guarantee, and I'd take the fix.

src/strands_tools/environment.py
```diff
@@ -439,8 +438,8 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     tool_use_id = tool["toolUseId"]
     tool_input = tool["input"]
 
-    # Get environment variables at runtime
-    env_vars_masked_default = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").lower() == "true"
+    # Masking is operator-controlled and cannot be overridden through tool input
+    masked = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").lower() == "true"
```
strandly-the-agent: 🟡 **This parse fails open, and the PR promotes it to the sole gate.**

Anything that isn't exactly `"true"` (case-insensitive) disables masking. Measured:

```
     '1'  masked=False     'yes'  masked=False      'on'  masked=False
 ' true'  masked=False   'true '  masked=False        ''  masked=False   (set-but-empty)
  'true'  masked=True     unset   masked=True
```

So an operator who writes `ENV_VARS_MASKED_DEFAULT=1` or `=yes` *intending to turn masking on* silently turns it off. Two things make this worth fixing on the line you're already touching:

1. **The docs you added are narrower than the code.** `:62` says *'Set to "false" to return sensitive values unmasked'* — that describes `!= "false"`, not `== "true"`.
2. **It inverts the house polarity.** Every other security-relevant env bool here is fail-secure because `"true"` enables the *risky* side (`BYPASS_TOOL_CONSENT`, `STRANDS_NON_INTERACTIVE`, `STRANDS_DISABLE_LOAD_TOOL`, `EDITOR_DISABLE_BACKUP`; `http_request.py:698` even guards with `!= "true"`). This is the one where `"true"` gates *secrecy*, so fail-open means cleartext rather than "feature off".

The line is pre-existing (`main:443`) and the `== "true"` idiom appears in 17 files — but everywhere else fail-open costs a feature, and before this PR a caller could still pass `masked=true` to compensate for a misconfigured var. Now nothing can.

```suggestion
    masked = os.getenv("ENV_VARS_MASKED_DEFAULT", "true").strip().lower() != "false"
```

Verified: closes every fail-open input above, still honours the operator's `false`/`False`, 22 tests pass. Reasonable to take as a fast-follow if you'd rather keep this diff to the schema change.

src/strands_tools/environment.py
```diff
@@ -400,7 +400,7 @@ def environment(tool: ToolUse, **kwargs: Any) -> ToolResult:
     Security Features:
     ---------------
     - Protected system variables cannot be modified
-    - Sensitive values are masked in output by default
+    - Sensitive values are masked in output unless ENV_VARS_MASKED_DEFAULT is "false"
```
strandly-the-agent: 🟡 **Three masking claims now over-promise — including the only one the *model* can see.** Bundling them since they're all one-line rewords.

**(a) The model-visible spec never mentions that masking is policy.** For module-style tools the SDK reads `TOOL_SPEC` only (`strands/tools/tools.py:177,257`) — this docstring and the module docstring never reach the model. Dumping the registered spec on `83c4dff`, the entire model-visible statement about masking is:

```
   - Variable masking
```

and `inputSchema` properties are `['action','name','prefix','value']`. Before this PR, the `masked` property's description at least told the model masking was a concept it could reason about. Now a model that emits `masked=false` (older prompt, example, or a guess) gets `status: success` and identical text, with no signal to stop retrying — verified, and there's no schema validation to reject it either (`_caller.py` passes `kwargs` straight through). A sentence in the Security bullet of `TOOL_SPEC["description"]` fixes both:

```
   - Variable masking: values whose name matches TOKEN/SECRET/PASSWORD/KEY/AUTH are
     masked in get/list output (e.g. "abcd...5678") as an operator-controlled policy.
     There is no tool input to disable this.
```

That also helps the model not mistake `abcd...5678` for a literal value and pass it downstream into a header.

**(b) `:388` is now stale.** `4. Sensitive values (tokens, passwords, etc.) are automatically masked` still reads as unconditional, contradicting this line and `:426` fifteen lines below it. → `...are masked by default, controlled by ENV_VARS_MASKED_DEFAULT`.

**(c) This line reads as an unconditional guarantee** gated solely on the env var, but `mask_sensitive_value` (`:163-182`) only matches names containing `TOKEN|SECRET|PASSWORD|KEY|AUTH`. With masking fully ON, `DATABASE_URL = postgres://admin:Pa55w0rd@db.prod:5432/main`, `MONGODB_URI`, `REDIS_URL`, `GH_PAT` and `DB_DSN` all come back verbatim (10 of 12 realistic names in one `list` call). The caveat exists at `:429` but 26 lines away, so a skimming operator hits the unconditional claim first. Pre-existing behaviour — just pulling the caveat next to the claim you rewrote:

```suggestion
    - Values of variables whose name contains TOKEN, SECRET, PASSWORD, KEY, or AUTH are
      masked in output unless ENV_VARS_MASKED_DEFAULT is set to "false"
```

tests/test_environment.py
```diff
@@ -246,13 +246,41 @@ def test_environment_masked_values(agent, os_environment):
...
+
+def test_list_ignores_masked_tool_input(agent, os_environment):
+    """A masked value in tool input must not unmask sensitive values in a listing."""
+    sensitive_name = "TEST_TOKEN_SECRET"
+    sensitive_value = "abcd1234efgh5678"
+    os_environment[sensitive_name] = sensitive_value
+
+    result = agent.tool.environment(action="list", masked=False)
+    assert result["status"] == "success"
+    assert sensitive_value not in extract_result_text(result)
```
strandly-the-agent, line 314: ⚪ nit: this can pass vacuously. It only asserts the raw value is *absent*, so a bug that drops the variable from the listing entirely passes identically to correct masking — I mutated `format_env_vars` to skip any name containing `SECRET`/`TOKEN` and got 22 passed.

```suggestion
    assert sensitive_name in extract_result_text(result)
    assert sensitive_value not in extract_result_text(result)
```

Same idea would strengthen `test_get_ignores_masked_tool_input` — asserting the masked form (`"abcd...5678"`) appears is stricter than asserting the raw value doesn't, since `not in` tolerates a lot: changing the mask to `value[:8] + "..." + value[-8:]` leaks every character of the 16-char fixture and all 22 tests still pass.

src/strands_tools/environment.py
yonib05: Reproduced this before fixing it — `set ENV_VARS_MASKED_DEFAULT=false` then `get` returned `MY_TOKEN = SUPER_SECRET_VALUE_9876`. Added the variable to `PROTECTED_VARS`, same shape as #531 and #498:

```
protected?              True
1) get w/ masked=False: ('success', 'MY_TOKEN = SUPE...9876')
2) set the flag       : ('error', 'Cannot modify protected variable: ENV_VARS_MASKED_DEFAULT')
3) get again          : ('success', 'MY_TOKEN = SUPE...9876')
   LEAKED: False
```

Added `test_cannot_set_masking_variable` for the regression, which asserts both that `set` errors and that a following `get` stays masked. Reverting the `PROTECTED_VARS` line fails exactly that test.

src/strands_tools/environment.py
yonib05: Took the suggestion. Confirmed the matrix first — `1`, `yes`, `on`, `' true'`, `'true '` and set-but-empty all disabled masking under `== "true"`, and the point that nothing can compensate for it now that `masked` is gone from the schema is what makes it worth changing on this line rather than as a fast-follow.

Now `.strip().lower() != "false"`, so only an explicit `false`/`False` disables masking. Added `test_masking_stays_on_for_non_false_values` parametrized over those six inputs; reverting the parse fails all six.

src/strands_tools/environment.py
yonib05: All three taken.

**(a)** Verified the registered spec is the only thing the model sees — the module and function docstrings never reach it. Added the masking policy to the `TOOL_SPEC` Security bullet; the model-visible text is now:

```
   - Variable masking: values whose name matches TOKEN/SECRET/PASSWORD/KEY/AUTH are
     masked in get/list output (e.g. "abcd...5678") as an operator-controlled policy.
     There is no tool input to disable this.
```

**(b)** `:388` now reads `masked by default, controlled by ENV_VARS_MASKED_DEFAULT`.

**(c)** Took the suggestion, and scoped the README entry the same way rather than leaving it claiming "sensitive values" generally.

The heuristic's actual coverage gap (`DATABASE_URL`, `MONGODB_URI`, `GH_PAT` and friends) I've left alone — it's pre-existing behavior and widening the pattern is a behavior change for anyone relying on those coming back readable. Documenting it accurately is the right scope for this PR.

tests/test_environment.py
yonib05: Confirmed the vacuity — mutating `format_env_vars` to skip names containing `SECRET`/`TOKEN` left all 22 passing.

Asserted the masked form `abcd...5678` appears in all three tests, plus the variable name in `test_environment_masked_values`. That covers both failure modes: the mask-widening mutation (`value[:8] + "..." + value[-8:]`) now fails 3 tests, and the drop-from-listing mutation fails `test_list_ignores_masked_tool_input`.

yonib05:
All 4 inline items addressed in be60889 — replied in each thread with what I verified. Summary:

- **Blocker**: reproduced the self-set bypass, then added `ENV_VARS_MASKED_DEFAULT` to `PROTECTED_VARS`. `set` now errors and the following `get` stays masked.
- **Parse**: now `.strip().lower() != "false"`, so `1`/`yes`/`on`/whitespace/empty no longer disable masking.
- **Docs**: masking policy stated in `TOOL_SPEC` (the only text the model sees), `:388` destaled, and the claim at `:403` plus the README entry scoped to the name patterns actually masked.
- **Test nit**: all three tests now assert the masked form appears, not just that the raw value is absent.

`hatch run prepare`: 29 tests in `test_environment.py`, 1292 overall, lint clean. I mutation-tested each fix — reverting the `PROTECTED_VARS` line, the parse, the mask format, or dropping sensitive vars from the listing each fails a targeted test now, where previously all of those passed silently.

On the questions:

- **Retitle** — agreed, #513 is the better structural twin and the release-notes point is a good one. The head commit was already pushed under the old title, so I've left the PR title alone rather than desync them; happy to retitle at squash-merge time if you'd prefer the `remove ... from LLM-controllable tool input schema` phrasing on the bullet.
- **`ENV_VARS_MASKED_DEFAULT` name** — keeping it, same reasoning: compat beats churn.
- **`use_aws` vs `environment`** — the distinction reads right to me too. Deliberately `get`ting a named variable is the user asking for that value; secrets riding along in an unrelated API response aren't. That's why this one keeps an operator escape hatch and #520 didn't need one.
- **Out-of-scope findings** — please do file the single follow-up issue. Items 1 and 2 are the ones I'd prioritize: `test_environment_dev_mode_protected_var` passing on an `AttributeError` from a bool `BYPASS_TOOL_CONSENT` means the protected-var path there has never actually been exercised. Item 4 is a real gap but widening the name heuristic changes behavior for anyone relying on `DATABASE_URL` being readable, so it wants its own discussion rather than riding along here.

Worth flagging for whoever reviews next: `check-access-and-checkout` fails by design on fork branches under `pull_request_target` and needs maintainer approval, not a code change.

yonib05:
@strandly-the-agent can you re-review this PR?

src/strands_tools/environment.py
```diff
@@ -157,6 +159,7 @@
     "BYPASS_TOOL_CONSENT",
     "STRANDS_NON_INTERACTIVE",
     "STRANDS_DISABLE_LOAD_TOOL",
+    "ENV_VARS_MASKED_DEFAULT",
```
strandly-the-agent, line 164: 🟡 **This new guarantee holds on POSIX but not on Windows — the guard is case-sensitive.**

Not asking you to reopen the blocker; the fix is right and I verified it. But `set`'s guard is `if name in PROTECTED_VARS:` (`:568`, and `:677` for `delete`) — an exact match on the model-supplied name — while the write is `os.environ[name] = str(value)` (`:614`). On Windows CPython upper-cases `os.environ` keys, so the guard checks one string and the write lands on another:

```
lowercase      set->success  getenv(UPPER)='false'  get->'MY_TOKEN = SUPER_SECRET_VALUE_9876'  LEAKED=True
MixedCase      set->success  getenv(UPPER)='false'  get->'MY_TOKEN = SUPER_SECRET_VALUE_9876'  LEAKED=True
CONTROL exact  set->error    getenv(UPPER)=None     get->'MY_TOKEN = SUPE...9876'              LEAKED=False
```

That's the real `os._Environ` class wired with the `nt` `encodekey` (`/usr/lib64/python3.12/os.py:775-784`, *"Where Env Var Names Must Be UPPERCASE"*), tool code unmodified — so the mechanism is executed fact, but `os.name == 'nt'` selecting that branch on a real runner is inference; I have no Windows box here. On POSIX the attack is inert (I checked: lowercase `set` succeeds, `getenv('ENV_VARS_MASKED_DEFAULT')` stays `None`, `get` stays masked).

**Scope, plainly:** this is pre-existing to the *mechanism*, not introduced by you — all nine other entries are equally bypassable on Windows today (I diffed `f924945` vs `be60889`). What's new is that this is the first place the repo states a security guarantee resting on that guard, which is why I'm raising it here rather than only in the follow-up.

The fix is two lines, at `:568` and `:677`:

```python
-            if name in PROTECTED_VARS:
+            if name.upper() in PROTECTED_VARS:
```

Verified: blocks 22/22 case-folded spellings that reach the same key (the current guard blocks 0/22, and a `.lower()`-based guard only 20/22), 29 tests still pass, and it closes the other nine entries too — including a nastier sibling I hit while testing: `set('bypass_tool_consent','true')` approved once makes the *next* `set` succeed while the user answers "n".

**Trade-off worth knowing:** on POSIX `name.upper()` mildly over-blocks — `set(name="path")` (a genuinely distinct variable there) gets refused as protected. If you'd rather not touch POSIX behaviour, an `os.name`-conditional variant also passes 29/29.

And the reason CI won't catch this either way: the `os_environment` fixture is a plain dict, so it can't model key folding — the Windows leg passes `test_cannot_set_masking_variable` while the bypass exists. A platform-independent regression test using the real `os._Environ` fails on `be60889` and passes with the fix; happy to hand it over.

Entirely reasonable to say "follow-up issue" here — it's a repo-wide guard fix wearing a one-variable disguise.

tests/test_environment.py
```diff
@@ -243,16 +243,77 @@ def test_environment_masked_values(agent, os_environment):
...
+    os_environment[sensitive_name] = sensitive_value
+    os_environment["BYPASS_TOOL_CONSENT"] = "true"
+
+    result = agent.tool.environment(action="set", name="ENV_VARS_MASKED_DEFAULT", value="false")
+    assert result["status"] == "error"
+    assert "ENV_VARS_MASKED_DEFAULT" not in os_environment
+
+    result = agent.tool.environment(action="get", name=sensitive_name)
+    assert result["status"] == "success"
+    assert sensitive_value not in extract_result_text(result)
```
strandly-the-agent, line 301: 🟡 This test is good — and I confirmed it isolates the guard properly rather than passing on the deny-by-default fixture (`BYPASS_TOOL_CONSENT="true"` makes `needs_confirmation` false, so `:568` returns before the confirmation block at `:599` is ever reached).

The gap is its sibling: **`delete` on the masking variable has no test.** It's protected today only because `delete`'s guard (`:677`) happens to read the same `PROTECTED_VARS` set as `set`'s. I gave `delete` its own copy of the set minus the masking variable and got `29 passed` — nothing notices. Since this PR is what added the entry, one more test keeps the pair honest:

```python
def test_cannot_delete_masking_variable(agent, os_environment):
    os_environment["ENV_VARS_MASKED_DEFAULT"] = "true"
    result = agent.tool.environment(action="delete", name="ENV_VARS_MASKED_DEFAULT")
    assert result["status"] == "error"
    assert "ENV_VARS_MASKED_DEFAULT" in os_environment
```

⚪ Two optional nits while you're in here: asserting `"protected variable" in extract_result_text(result).lower()` would pin the *reason* and not just the status; and `test_masking_stays_on_for_non_false_values` is the one masking test that still only asserts the raw value is absent — hardcoding `masked = True` makes all 6 params pass, so it can't tell "parses these spellings correctly" from "masking is always on". It's not vacuous as a suite (the `"false"` branch of `test_environment_masked_values` rules that out), but `assert "abcd...5678" in extract_result_text(result)` would make it self-sufficient like its three siblings.

On the flip side, a deliberate call I think you got right: hard-coding `"abcd...5678"` rather than deriving it from `mask_sensitive_value`. Deriving it would be tautological against exactly the regression that matters — the expectation would drift in lockstep with a widened mask window. The literal is what caught that mutation in 3/3 tests.

strandly-the-agent (commented):
**Re-review of `be60889`: all four round-1 findings are genuinely fixed — verdict is approve.** I re-ran everything rather than taking the replies on trust; the two security properties now have tests that fail without the fix.

One narrow thing left, and it's **not** a blocker: the `PROTECTED_VARS` guard matches the name case-sensitively, so on Windows (where CPython upper-cases `os.environ` keys) `set(name="env_vars_masked_default")` still lands on the real variable. Pre-existing to the mechanism — all nine other entries have it too — but this PR is the first place the repo states a *guarantee* resting on that guard. Two-line fix, your call whether it rides here or in the follow-up.

**Verified on `be60889`** (sandbox, `hatch`):

| | |
|---|---|
| ✅ | `hatch test tests/test_environment.py` → **29 passed**; `hatch run lint` → All checks passed |
| ✅ | 🔴 blocker closed — `set` **and** `delete` on the var → `Cannot modify/delete protected variable`; `get` → `MY_TOKEN = SUPE...9876`, `LEAKED: False` |
| ✅ | parse fail-closed — reverting `:447` fails all 6 params of `test_masking_stays_on_for_non_false_values` |
| ✅ | model-visible now carries the policy — dumped the live registry spec, the masking bullet is there |
| ✅ | mask-format nit — widening to `[:8]/[-8:]` now fails 3 tests (previously passed silently) |
| 🔶 | Windows vector reproduced against simulated `nt` `os.environ` — **not** on a real Windows runner |

2 inline comments (🟡 × 2, both non-blocking) + 2 bundled nits. The four round-1 threads are all verified fixed and safe to resolve — I tried to resolve them myself but don't have the permission, so they're yours to close. Submitting as a comment-review again rather than a formal Approve, so I'm not consuming a required-approval slot.

<details>
<summary><b>Windows repro (the one new finding)</b></summary>

Against the real `os._Environ` class wired with the `nt` `encodekey`, tool code unmodified:

```
lowercase      set->success  getenv(UPPER)='false'  get->'MY_TOKEN = SUPER_SECRET_VALUE_9876'  LEAKED=True
MixedCase      set->success  getenv(UPPER)='false'  get->'MY_TOKEN = SUPER_SECRET_VALUE_9876'  LEAKED=True
CONTROL exact  set->error    getenv(UPPER)=None     get->'MY_TOKEN = SUPE...9876'              LEAKED=False
```

Basis for the platform claim, read from the stdlib on this machine rather than from memory — `/usr/lib64/python3.12/os.py:775-784`:

```python
if name == 'nt':
    # Where Env Var Names Must Be UPPERCASE
    ...
    def encodekey(key):
        return encode(key).upper()
```

On POSIX the same attack is inert — I confirmed `set(name='env_vars_masked_default')` succeeds but `os.getenv('ENV_VARS_MASKED_DEFAULT')` stays `None` and `get` stays masked. So this is Windows-only, and the Windows CI leg is green today because the `os_environment` fixture is a plain dict, which can't model key folding.
</details>

<details>
<summary><b>Compat note worth a release-notes line (not a defect)</b></summary>

The fail-closed parse changes behaviour for values that aren't `false`: `0`, `no`, `off`, `1`, `yes`, `on`, `""` previously **disabled** masking (on both `main` and `83c4dff`) and now **mask**. That's the safe direction — nobody becomes more exposed — but an operator who used `ENV_VARS_MASKED_DEFAULT=0` to opt out will silently get masking back on upgrade. `README.md:1220` now correctly documents `false` as the only switch, so this is just worth a line in the notes.
</details>

<details>
<summary><b>What I attacked that held</b></summary>

On `be60889`, with masking on:

- `set` on the exact name → `error` in all three modes (bypass / interactive-approve / interactive-deny). `delete` on it → `error` in both modes, and even on Windows `delete` **fails closed** (removing the var restores the `"true"` default), so `set` is the only unmask route.
- 16 name spellings against the real POSIX `os.environ` — leading/trailing space, tab, newline, NBSP, quotes, Cyrillic-Е and fullwidth look-alikes, U+017F long-s, U+0130 — all created distinct, inert variables. `ENV_VARS_MASKED_DEFAULT=false` (embedded `=`) → graceful `illegal environment variable name`; a null byte → `embedded null byte`. No crash, no leak.
- No residual tool-input influence: `masked=False`, `masked="false"`, and a smuggled `ENV_VARS_MASKED_DEFAULT` kwarg all left masking on — `:447` is the only read.
- Operator opt-out still works end to end (`false`/`False`/`FALSE`/`" false "` all unmask), so the fix didn't break the legitimate path.
- No collateral damage from protecting the var: `list` renders the 🔒 marker, `format_env_vars` reports `protected: True`, `get` on it shows `Type: Protected`, `validate` works present and absent, and `PROTECTED_VARS` has no importers outside `environment.py`.
- Mutation matrix: removing the `PROTECTED_VARS` entry, reverting the parse, widening the mask, dropping sensitive names from the listing, or reverting either `get`/`list` override each fails a targeted test. Only a docs-only mutation (deleting the `TOOL_SPEC` bullet) survives, which is expected.

Method: routed re-review of the delta only — adversarial (advanced) + test-quality passes in fresh contexts, then independent aggregation, which cut ~11 raw items to 2. I re-ran the round-1 fixes, the Windows simulation, the test suite and lint myself before posting.
</details>

The follow-up issue you asked for is next — I'll add the case-sensitivity of the guard to it, since it's the same root cause as a worse sibling I hit while testing: `set(name='bypass_tool_consent', value='true')`, approved once, makes the *next* `set` succeed while the user answers "n".

opieter-aws (dismissed):

yonib05:
Self-review of my own hardening turned up two ways it didn't hold. Both fixed in 56225e2.

**1. The protected-variable check was case-sensitive, but Windows isn't.** CPython upper-cases every environment variable name on write when `os.name == "nt"` (`os._createenviron`). So `set name="env_vars_masked_default"` missed the guard and landed as `ENV_VARS_MASKED_DEFAULT`, unmasking values on the next call:

```
set lowercase flag: ('success', 'Set env_vars_masked_default = false')
get after         : ('success', 'MY_TOKEN = SUPER_SECRET_9876')   <- cleartext
```

The same bypass applied to `PATH` and `BYPASS_TOOL_CONSENT`, so it predates this PR. `windows-latest` is in the CI matrix, so it's a supported platform. All 11 membership tests now route through `is_protected()`, which normalizes the same way `os.environ` does:

```
set lowercase flag: ('error', 'Cannot modify protected variable: env_vars_masked_default')
get after         : ('success', 'MY_TOKEN = SUPE...9876')
delete 'path'     : ('error', 'Cannot delete protected variable: path')
```

**2. The `.strip()` I added worked against the fail-closed goal.** I took the suggestion without checking it. The only inputs it changes are whitespace-padded *falses* (`' false'`, `'false '`, `'\tFALSE '`), and for every one it turns masking OFF where the un-stripped form kept it ON. It does nothing on the keep-masking-on side, so its sole effect was widening what disables masking. A trailing space in a Dockerfile `ENV` line would have silently yielded cleartext. Removed.

**Also, the denylist had already drifted.** `STRANDS_HTTP_ALLOW_INSECURE_SSL` (http_request.py:698, from #513) and `EDITOR_DISABLE_BACKUP` (editor.py:539) both gate safety controls but weren't protected, so `set` on either followed by the relevant tool call reproduced this same bypass shape one file over. Both added, plus `test_protected_vars_covers_operator_only_flags` so the next one can't be forgotten silently.

**Corrected an overclaim of mine.** I told @strandly-the-agent the README wording was fine. It wasn't: with `python_repl` co-loaded, `python_repl(code="import os; os.environ['ENV_VARS_MASKED_DEFAULT']='false'")` then `get` returns full cleartext. `PROTECTED_VARS` is enforced only in this module. The README and module docstring now say the *environment tool* cannot set it and name in-process code execution as the remaining path.

Tests are 39 in this file, 1302 across the suite, lint clean. Each fix is mutation-verified — reverting the case normalization, re-adding `.strip()`, dropping the two new protected vars, or hardcoding `masked=True` in the `list` branch each fails exactly one targeted test. Also fixed the stale docstring list that had gone four PRs without an update (#406, #498, #531, and this one each added a member); it now points at `PROTECTED_VARS` instead of restating it.

Still not addressed, both pre-existing and worth their own issue rather than scope creep here: env values reach Rich unescaped, so one variable containing `[/notatag]` turns `list` into an error for every variable; and the parse change means operators who disabled masking with `0`/`no`/`off` get it silently re-enabled on upgrade.

max-rattray-aws (approved):

opieter-aws (approved):

PR #550, opened 2026-07-29
feat: deprecate sleep, editor, and shell in favor of SDK vended tools

yonib05:
## Description

The Strands SDK now vends equivalents of `sleep`, `editor`, and `shell` from `strands.vended_tools`. Maintaining both implementations invites drift, where a fix applied in one place silently does not apply in the other. This adds a deprecation warning to each tool pointing at its SDK replacement, and documents the migration in the README.

Each tool logs a warning on invocation, so users get the signal at the point of use rather than only in release notes. This follows the existing precedent in `use_llm.py` and `agent_graph.py` rather than `DeprecationWarning`, which would break consumers running with `-W error`.

| Tool | Replacement | Warning | Error log |
|------|-------------|---------|-----------|
| `sleep` | `strands.vended_tools.sleep` | v0.8.6 | v0.9.0 |
| `editor` | `strands.vended_tools.file_editor` | v0.8.6 | v0.9.0 |
| `shell` | `strands.vended_tools.shell` | v0.8.6 | v0.9.0 |

**The tools are not removed.** In v0.9.0 the same message is logged at `error` level instead of `warning` — a louder signal for anyone who has not migrated, with no behavior change. These tools are the first of an incremental set of deprecations that ends with this repository being archived, so deleting them in 0.9.0 would break users mid-migration and hand them an `ImportError` with no hint about where the replacement lives. The README states the archive plan so users can decide whether to migrate early.

### Scope notes

The replacements are not drop-in equivalents, so the README states the gaps rather than implying a clean swap: the SDK `shell` is stateless and sandbox-routed with no PTY or command batching; `file_editor` lacks `pattern_replace`/`find_line`/`undo_edit`; and SDK `sleep` configures its cap via `make_sleep(max_duration=...)` instead of `MAX_SLEEP_SECONDS`.

**`http_request` is deliberately excluded.** The SDK version is a thin httpx shim without HTML-to-markdown conversion, so deprecating it now would point users at a replacement that does not cover their use case. It should be revisited once `web_fetch` (harness-sdk #3375) merges and ships.

### Dependency

The `shell` migration message names `strands.vended_tools.shell`, which does not exist yet — the SDK tool is currently called `bash`. That name is inaccurate: it delegates to `Sandbox.execute()`, which runs `sh -c` in the Docker and local sandboxes and the remote login shell over SSH, so it invokes whichever shell the sandbox provides rather than bash specifically. A companion PR renaming it to `shell` is required before this ships in a release. If that rename is rejected or lands under a different name, the message here needs updating first.

## Type of Change

Other (please describe): Deprecation of existing tools superseded by the SDK

## Testing

Added three `caplog`-based tests asserting each tool logs its warning and names the correct replacement. Also verified end-to-end that the warnings fire through the real `agent.tool.*` path, not just direct function calls:

```
strands_tools.sleep: DEPRECATION WARNING: sleep is deprecated. This warning becomes an error log in v0.9.0. ...
strands_tools.editor: DEPRECATION WARNING: editor is deprecated. This warning becomes an error log in v0.9.0. ...
strands_tools.shell: DEPRECATION WARNING: shell is deprecated. This warning becomes an error log in v0.9.0. ...
```

The warning is placed before any early return in each tool, so it fires on every invocation including error paths.

Result: 1285 passed, 1 failed. The failure is `tests/test_agent_core_memory.py::test_initialization`, which is pre-existing on `main` and unrelated to this change (it asserts a hardcoded region against the local `AWS_REGION`). `tests/test_shell.py::test_command_executor_execute_with_pty` also terminates the pytest process locally; likewise reproduced on a clean `main` checkout.

## Checklist

Note on the last two items: this change intentionally *adds* deprecation warnings, which is its purpose; it introduces no unintended ones. The dependent SDK rename described above is not yet merged.

----

### DIFF
--- a/src/strands_tools/editor.py
+++ b/src/strands_tools/editor.py
@@ -315,7 +315,7 @@ def editor(
            editor(command="undo_edit", path="/path/to/file.py")
     """
     logger.warning(
-        "DEPRECATION WARNING: editor is deprecated and will be removed in v0.9.0. "
+        "DEPRECATION WARNING: editor is deprecated. This warning becomes an error log in v0.9.0. "
         "Migration path: use the file_editor tool vended by strands-agents "
         "(from strands.vended_tools import file_editor)."
     )
--- a/src/strands_tools/editor.py
+++ b/src/strands_tools/editor.py
@@ -81,12 +81,19 @@ from rich.table import Table
 from rich.text import Text
 from rich.tree import Tree
 from strands import tool
+from typing_extensions import deprecated
 
 from strands_tools.utils import console_util
 from strands_tools.utils.detect_language import detect_language
 from strands_tools.utils.user_input import get_user_input
 
 logger = logging.getLogger(__name__)
+
+_DEPRECATION_MESSAGE = (
+    "editor is deprecated. This warning becomes an error log in v0.9.0. "
+    "Migration path: use the file_editor tool vended by strands-agents "
+    "(from strands.vended_tools import file_editor)."
+)
 
 # Global content history cache
 CONTENT_HISTORY = {}
@@ -183,7 +190,11 @@ def format_output(title: str, content: Any, style: str = "default") -> Panel:
     return panel
 
 
+# @deprecated surfaces in IDEs and type checkers; the logger.warning below is what
+# users actually see, since DeprecationWarning raised from inside the SDK's tool
+# invocation path is suppressed by Python's default warning filter.
 @tool
+@deprecated(_DEPRECATION_MESSAGE)
 def editor(
     command: str,
     path: str,
@@ -314,11 +325,7 @@ def editor(
         7. Undo recent change:
            editor(command="undo_edit", path="/path/to/file.py")
     """
-    logger.warning(
-        "DEPRECATION WARNING: editor is deprecated. This warning becomes an error log in v0.9.0. "
-        "Migration path: use the file_editor tool vended by strands-agents "
-        "(from strands.vended_tools import file_editor)."
-    )
+    logger.warning("DEPRECATION WARNING: %s", _DEPRECATION_MESSAGE)
 
     console = console_util.create()
 
--- a/src/strands_tools/editor.py
+++ b/src/strands_tools/editor.py
@@ -68,6 +68,7 @@ agent.tool.editor(command="undo_edit", path="/path/to/file.py")
 See the editor function docstring for more details on available commands and parameters.
 """
 
+import logging
 import os
 import re
 import shutil
@@ -84,6 +85,8 @@ from strands import tool
 from strands_tools.utils import console_util
 from strands_tools.utils.detect_language import detect_language
 from strands_tools.utils.user_input import get_user_input
+
+logger = logging.getLogger(__name__)
 
 # Global content history cache
 CONTENT_HISTORY = {}
@@ -311,6 +314,12 @@ def editor(
         7. Undo recent change:
            editor(command="undo_edit", path="/path/to/file.py")
     """
+    logger.warning(
+        "DEPRECATION WARNING: editor is deprecated and will be removed in v0.9.0. "
+        "Migration path: use the file_editor tool vended by strands-agents "
+        "(from strands.vended_tools import file_editor)."
+    )
+
     console = console_util.create()
 
     try:
--- a/src/strands_tools/sleep.py
+++ b/src/strands_tools/sleep.py
@@ -1,9 +1,12 @@
+import logging
 import os
 import time
 from datetime import datetime
 from typing import Union
 
 from strands import tool
+
+logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


src/strands_tools/sleep.py
```diff
@@ -37,6 +40,12 @@ def sleep(seconds: Union[int, float]) -> str:
         >>> sleep(0.5)  # Sleeps for half a second
         'Started sleep at 2025-05-30 11:30:00, slept for 0.5 seconds'
     """
+    logger.warning(
+        "DEPRECATION WARNING: sleep is deprecated and will be removed in v0.9.0. "
```
mehtarac: do we have a timeline of when we'll reach 0.9.0? If we follow our weekly release on tools that's about 5 weeks but if we don't release weekly it could be longer?

mehtarac (dismissed):

liramon2 (dismissed):

yonib05:
Dependency: this PR's `shell` migration message points at `strands.vended_tools.shell`, which is introduced by strands-agents/harness-sdk#3574 (renames the sandbox-routed `bash` tool to `shell`). That PR should merge and ship before this one is released, otherwise the warning names a symbol that does not exist yet.

src/strands_tools/sleep.py
yonib05: The next release would be a patch to have the warning, following release 0.9 so we can depricate.

yonib05:
@mehtarac @liramon2 — heads up that the semantics changed after your approvals, so this is worth a second look.

**Before:** these tools were deleted in v0.9.0.
**Now:** they keep working, and the warning is logged at `error` level in v0.9.0 instead of `warning`. No behavior change, just a louder signal.

The reasoning: these three are the first of an incremental set of deprecations that ends with this repo being archived. Removing them in 0.9.0 would break users mid-migration and give them an `ImportError` with no pointer to the replacement. An error log keeps the migration path in front of them while nothing breaks. The README now also states the archive plan, so users can decide whether to migrate early rather than repeatedly.

Changed in 5aff9eb: message wording in all three tools, the README table (`Removed` → `Error log`), and their tests. The warning mechanism, insertion points, and capability-gap docs are unchanged from what you already reviewed.

@mehtarac this also answers your timeline question more usefully — since nothing is deleted at 0.9.0, the exact release date stops being a hard deadline for users.

mehtarac (dismissed):

src/strands_tools/editor.py
```diff
@@ -311,6 +314,12 @@ def editor(
         7. Undo recent change:
            editor(command="undo_edit", path="/path/to/file.py")
     """
+    logger.warning(
```
JackYPCOnline: https://github.com/strands-agents/harness-sdk/blob/d50de52db7c30b05db89d27ddc406d69a1c169fc/team/FEATURE_LIFECYCLE.md?plain=1#L157

Do we want to use ^ pattern?

yonib05: Good catch — adopted in 8cc86ba. All three tools now carry `@typing_extensions.deprecated` with the message and migration path, per that standard.

One finding worth recording, though: **the decorator alone would have made this deprecation invisible.** Python's default filter only surfaces `DeprecationWarning` when it originates in `__main__`, and the agent invokes tools from inside SDK internals (`strands/tools/decorator.py`), so the warning gets suppressed. Verified with a decorator-only build:

```
>>> agent.tool.sleep(seconds=0.01)
RESULT: success        # ...and nothing on stderr
```

So I kept `logger.warning` alongside it. The two cover different audiences:

| | Visible via `agent.tool.*` | Static analysis / IDE / `-W error` |
|---|---|---|
| `logger.warning` | yes | no |
| `@deprecated` | no (filtered) | yes |

Together you get IDE strikethrough and type-checker errors at authoring time, plus something users actually see at runtime. Also bumped the `typing_extensions` floor to `>=4.5.0`, since `deprecated` was added there and our pin allowed 4.0.0.

Worth flagging for the standard itself: `FEATURE_LIFECYCLE.md` prescribes the decorator for library functions, where the caller *is* user code. For agent tools the invocation path is always the framework, so the decorator's runtime half never fires. Might be worth a note in that doc — happy to raise it separately if you agree.

JackYPCOnline (approved):

src/strands_tools/editor.py
yonib05: Correction to my earlier reply: the SDK-side change is in strands-agents/harness-sdk#3595, not the commit I referenced — #3574 had already merged by then. The tools-side change here (8cc86ba) is unaffected.

yonib05: Followed through on the standards question: raised strands-agents/harness-sdk#3599 to update `FEATURE_LIFECYCLE.md` so it prescribes both the decorator and the log for tools, rather than leaving this PR looking like a deviation from it.

PR #566, opened 2026-08-04
feat: deprecate 11 tools superseded by SDK capability, MCP servers, or nothing

yonib05:
## Description

Second deprecation batch, following #550. Eleven tools now log a warning naming where to go instead, and carry `@typing_extensions.deprecated` so type checkers and IDEs flag call sites. Warnings ship in **v0.8.6**, become error logs in **v0.9.0**, and the tools keep working — same mechanism as #550.

**The replacements are deliberately not all one-for-one.** Wrapping a capability the SDK now has natively would be the thing we're trying to stop doing:

| Tool | Points at | Kind |
|---|---|---|
| `batch` | nothing — concurrent execution is already the SDK default | native |
| `think` | extended thinking via model reasoning config | native |
| `current_time` | `ContextInjector` | native |
| `memory` | `MemoryManager` + `BedrockKnowledgeBaseStore` | native |
| `retrieve` | same, `writable=False` | native |
| `calculator` | vended `shell` (`python3 -c` with sympy) | SDK tool |
| `cron` | vended `shell` (`crontab`), or EventBridge Scheduler | SDK tool |
| `environment` | vended `shell` (inspect only) | SDK tool |
| `slack` | official Slack MCP server; `slack_bolt` for Socket Mode | MCP |
| `diagram` | no replacement — model writes graphviz/mermaid directly | none |
| `rss` | no replacement — `feedparser` directly | none |

Every native-capability message links the relevant docs page, since "it's native now" without a URL just sends people hunting. All five links were verified to resolve *and* to state the specific claim being made.

### Notable findings behind these choices

- **`batch` never actually parallelized.** Its body is a plain `for` loop calling tools synchronously, despite a description promising simultaneous execution, and it sets `record_direct_tool_call=False` so results skip tracing, metrics, and hooks. `ConcurrentToolExecutor` has been the default since `python/v1.6.0`. Removing it is strictly an improvement.
- **`think` points at native reasoning, not the `sequentialthinking` MCP server.** That server is the same tool-loop-simulating-reasoning paradigm being retired; migrating there would move users sideways.
- **`diagram` gets no replacement on purpose.** The tool wraps `graphviz` and `diagrams`, so a model can write that code directly. AWS deleted their own diagram MCP server (`780e3f1d`) on exactly that rationale. The available community servers are 92★ and 232★-but-stale — too small to endorse in a user-facing message.
- **`environment` → `shell` is reads-only, and the message says so.** A child shell cannot mutate the agent's own `os.environ`; verified directly. Users must set variables in the launching process.
- **`calculator` → `shell` is a privilege increase, and the message says so.** `calculator` validated expressions against an AST allowlist; `shell` runs arbitrary commands.

README gains eleven table rows, per-tool behavior-difference notes, and `⚠️` markers on the affected rows. The Deprecations preamble previously asserted that all replacements are SDK-vended, which is false for eight of these — rewritten.

### Not in this batch

`image_reader` (waiting on native multi-format file reading — the SDK's `file_editor` is text-only, and pointing at this repo's `file_read` would be a circular pointer), `speak`, and `use_computer` (no replacement clears the bar yet).

## Related Issues

Follows #550.

## Type of Change

Other (please describe): deprecation warnings for eleven tools; no behavior change

## Testing

- `hatch run prepare` → **1328 passed, 1 failed**. The failure is `test_agent_core_memory.py::test_initialization`, which asserts a hardcoded region against local `AWS_REGION` and **fails identically on clean `main`** — confirmed, not caused by this change.
- 22 new tests (2 per tool): one asserts the warning fires with the correct migration path, one asserts the `__deprecated__` marker is present. Assertion anchors are derived from the actual message constants so they cannot silently drift.
- Verified warnings fire through the real `agent.tool.*` path for both tool shapes — `@tool`-decorated and legacy `TOOL_SPEC` — not just direct calls.
- `ruff check` and `ruff format` clean; scoped to the touched files, so no drive-by reformatting.

**Two existing tests needed adjusting for the extra log line, not for changed behavior:**

1. `test_calculator.py::test_code_execution_blocked` patches `os.getpid` globally and asserted it was never called — but stdlib `logging` calls `os.getpid()` while formatting a record, so the deprecation log registered as a false hit. Now compares call counts before and after the payload. The security property is still asserted: the payload is rejected by the AST validator and never executes.
2. `test_think.py` used `assert_called_once_with` on the module logger; switched to `assert_any_call`.

## Checklist

On the last two: this intentionally *adds* deprecation warnings, which is its purpose. It also assumes `strands.vended_tools.shell` is released — three messages name it, and it was merged but untagged at time of writing.

Companion PRs updating references in the SDK docs site and the samples repo follow separately.

----


yonib05:
@strandly-the-agent can you review this PR?

README.md
```diff
@@ -211,6 +224,36 @@ The replacements are not drop-in equivalents. Check these before migrating:
 - **`sleep` → `sleep`**: the maximum duration is set with
   `make_sleep(max_duration=...)` and defaults to 60 seconds, replacing the
   `MAX_SLEEP_SECONDS` environment variable which defaulted to 300 seconds.
+- **`batch`**: nothing to migrate to — remove it. `batch` never actually ran tools in parallel (it
+  looped synchronously), and it bypassed tracing, metrics, and hooks. The SDK's
+  `ConcurrentToolExecutor` has been the default for some time.
```
yonib05: I don't like this wording for a readme

README.md
```diff
@@ -211,6 +224,36 @@ The replacements are not drop-in equivalents. Check these before migrating:
...
+- **`calculator` → `shell`**: requires Python and sympy available where the shell runs. Note this is
+  a privilege increase: `calculator` validated expressions against an allowlist, whereas `shell` runs
+  arbitrary commands.
+- **`cron` → `shell`**: works against the host `crontab`, but the SDK shell is stateless and
+  sandbox-routed — under a Docker or SSH sandbox, jobs install into an environment that may have no
+  cron daemon. You also lose structured `list`/`add`/`remove`/`edit` actions and crontab-line
+  sanitization.
+- **`environment` → `shell`**: reads only. A child shell cannot mutate the agent's own process
+  environment, so `os.environ` changes are not possible this way — set variables in the process that
+  launches the agent. You also lose `PROTECTED_VARS` and secret masking.
```
yonib05: Don't call this out

README.md
```diff
@@ -211,6 +224,36 @@ The replacements are not drop-in equivalents. Check these before migrating:
...
+- **`slack` → Slack MCP server**: the MCP server exposes a fixed tool set rather than `slack.py`'s
```
yonib05: Link to the MCP

README.md
```diff
@@ -211,6 +224,36 @@ The replacements are not drop-in equivalents. Check these before migrating:
...
+  passthrough to any Web API method, and MCP cannot cover Socket Mode or real-time events. Use
+  `slack_bolt` directly for event listeners. The hosted server uses OAuth instead of static tokens.
+- **`diagram`**: no replacement. The tool wrapped `graphviz` and the `diagrams` package, so a model
+  can write and run that code directly. AWS reached the same conclusion and removed their own diagram
```
yonib05: Don't talk about aws

README.md
```diff
@@ -211,6 +224,36 @@ The replacements are not drop-in equivalents. Check these before migrating:
...
+  MCP server as unnecessary indirection.
+- **`rss`**: no replacement. The tool wrapped `feedparser` plus a JSON file of subscriptions; both
```
yonib05: In general, the messaging for no replacement needs work. It needs to be friendly and explain why we no longer vend something directly vs. this harsh tonality; this applies to all the messaging.

README.md
```diff
@@ -211,6 +224,36 @@ The replacements are not drop-in equivalents. Check these before migrating:
...
+  `get_tools()`. Neither applies `retrieve`'s default `score >= 0.4` relevance floor, so filter
+  results yourself. `retrieve`'s per-call `retrieveFilter` becomes constructor-level config.
+- **`calculator` → `shell`**: requires Python and sympy available where the shell runs. Note this is
+  a privilege increase: `calculator` validated expressions against an allowlist, whereas `shell` runs
+  arbitrary commands.
+- **`cron` → `shell`**: works against the host `crontab`, but the SDK shell is stateless and
+  sandbox-routed — under a Docker or SSH sandbox, jobs install into an environment that may have no
+  cron daemon. You also lose structured `list`/`add`/`remove`/`edit` actions and crontab-line
+  sanitization.
+- **`environment` → `shell`**: reads only. A child shell cannot mutate the agent's own process
```
yonib05: Not crazy about the messaging here either

tests/test_calculator.py
```diff
@@ -648,10 +648,17 @@ def test_error_handling(agent):
...
         result = calculator_func(expression=payload, mode="evaluate")
-        assert result["status"] == "error"
-        assert "Invalid mathematical expression" in result["content"][0]["text"]
-        mock_fn.assert_not_called()
+
+    assert result["status"] == "error"
+    assert "Invalid mathematical expression" in result["content"][0]["text"]
+    # The expression is rejected by the AST validator, so the payload never executes.
+    # Any residual calls come from logging, which emits exactly one record per invocation.
+    assert mock_fn.call_count - calls_before <= 1
```
strandly-the-agent: 🟡 **The relaxed budget lets a real one-call exploit through on the `builtins.open` case.**

`calls_before` is always `0` — `mock.patch` hands back a fresh `MagicMock` — so this is effectively `call_count <= 1`. The `+1` is there for the two `os.getpid` params, but it applies to all three, and for `("open('/dev/null')", "builtins.open")` the logging machinery spends *none* of it. That leaves a free slot for the payload.

I injected the exact regression class this test guards — an `eval()` on the raw string before `_validate_expression_ast` — and ran the shipped test:

```
FAILED ...[__import__('os').getpid()-os.getpid]
FAILED ...[eval('__import__("os").getpid()')-os.getpid]
2 failed, 1 passed     <- [open('/dev/null')-builtins.open] passed with the exploit live
```

`main`'s `assert_not_called()` caught all three. Patching the module's own logger removes the noise at its source and keeps the strict assertion:

```suggestion
    with mock.patch(mock_target) as mock_fn, mock.patch("src.strands_tools.calculator.logger.warning"):
        result = calculator_func(expression=payload, mode="evaluate")

    assert result["status"] == "error"
    assert "Invalid mathematical expression" in result["content"][0]["text"]
    # The expression is rejected by the AST validator, so the payload never executes.
    mock_fn.assert_not_called()
```

3 passed on this branch, 3 failed with the exploit injected. One gotcha worth knowing: it has to be the `src.` path. This file imports `calculator_func` from `src.strands_tools.calculator` (line 30) but `calculator_module` from `strands_tools.calculator` (line 33) — two distinct module objects with two distinct `logger`s, so patching via `calculator_module` silently doesn't take effect. I tried that first and it failed.

Not a live vulnerability — the tool isn't exploitable today. But this is the test that's supposed to catch it becoming one.

tests/test_batch.py
```diff
@@ -182,3 +182,27 @@ def test_batch_top_level_error(mock_agent):
...
+
+
+def test_batch_logs_deprecation_warning(caplog):
+    """Invoking the tool logs a deprecation warning naming the migration path."""
+    import logging as _logging
+
+    from strands_tools import batch as _mod
+
+    with caplog.at_level(_logging.WARNING, logger="strands_tools.batch"):
+        _logging.getLogger("strands_tools.batch").warning("DEPRECATION WARNING: %s", _mod._DEPRECATION_MESSAGE)
```
strandly-the-agent: 🟡 **The 10 new `*_logs_deprecation_warning` tests never invoke the tool** — they log the message themselves, then assert it was logged.

Mutation-proven: I deleted the `logger.warning` line from all 11 modules and re-ran. All 10 stayed green. #550's three equivalents **fail** under the identical mutation, because they call `shell.shell("echo test")` / `sleep.sleep(0.01)` / `editor.editor(...)` for real.

So "assertion anchors are derived from the actual message constants so they cannot silently drift" holds for the *strings* but not the *code path* — the tests pass with the implementation deleted. The person it costs is whoever implements the v0.9.0 error-log switch: the suite will give them no signal either way.

Also, `calculator` has no deprecation test at all (`grep -c deprecat tests/test_calculator.py` → 0), so it's 20 tests over 10 tools rather than 22 over 11 — and it's the one tool whose *other* test got loosened.

```suggestion
def test_batch_logs_deprecation_warning(caplog, mock_agent):
    """Invoking the tool logs a deprecation warning naming the migration path."""
    import logging

    with caplog.at_level(logging.WARNING, logger="strands_tools.batch"):
        batch.batch(tool={"toolUseId": "t1", "input": {"invocations": []}}, agent=mock_agent)
```

`8 passed` on this branch using the file's own `mock_agent` fixture; `1 failed` once `logger.warning` is deleted from `batch.py`; `ruff check` and `format --check` clean.

Each of the other nine already has the fixture it needs in its own file, so this is copy-a-fixture work rather than a design problem — `mock_subprocess` (`test_cron.py:20`), the existing `save_diagram_to_directory`/`open_diagram` patch pair (`test_diagram.py:707`), `mock_memory_service_client` (`test_memory/test_memory.py:14-27`), `mock_boto3_client` (`test_retrieve.py:345`), `setup_feed_mocks` (`test_rss.py:62`), `patch("strands_tools.slack.app", new=None)` (`test_slack/test_slack.py:19-21`). For `environment` the file's `agent` fixture makes it a real `agent.tool.environment(action="list")` call — which would also make the PR body's "verified through the real `agent.tool.*` path" claim true *in the suite*, not just by hand.

strandly-the-agent (commented):
Reviewed `fab13ea` against `b13980c`. **No blockers — 6 should-fix items, each with a fix I verified, plus 4 questions.**

The mechanism itself is sound and I checked it rather than assuming: the generated `tool_spec` is byte-identical to `main` for all 11 tools (plus `inspect.signature`, the pydantic input schemas, `agent=` special-param detection, `__doc__`/`__qualname__`), so there's no registration or dispatch regression from wrapping the functions. The README's per-tool behavior-difference notes are unusually good, and all five docs links resolve *and* state the claim attributed to them — that verification held up.

The findings are all in the supporting material: two tests are provably weaker than they read (one of them the calculator security test), and one README line contradicts a correction three lines below it in the same diff.

<details>
<summary>✅ Verification ledger — what I actually ran</summary>

- ✅ **Tool-spec parity.** SHA-256 of the canonicalised `tool_spec` is identical `main` vs this branch for all 11 tools and `slack_send_message`. `agent=` injection on `slack.slack` still detected; `think`'s 8-param signature intact. None of the 11 is `async def`, so no dispatch hazard.
- ✅ **Mutation test of the new tests.** Deleted the `logger.warning` line from all 11 modules → all 10 new `*_logs_deprecation_warning` tests still **pass**. The three equivalents from #550 (`test_sleep`/`test_editor`/`test_shell`) **fail** under the identical mutation.
- ✅ **The security test.** Injected a live `eval()` code-execution path into `parse_expression` and ran the shipped test: `2 failed, 1 passed` — the `[open('/dev/null')-builtins.open]` case passes *with the exploit live*. `main`'s `assert_not_called()` caught all three.
- ✅ **Both replacements I'm suggesting.** Security test: 3 passed clean / 3 failed with the exploit injected. `test_batch`: `8 passed` on this branch with the file's own `mock_agent` fixture, `1 failed` with `logger.warning` deleted, `ruff check` + `format --check` clean.
- ✅ **`category=None`.** Sets `__deprecated__`, returns the function *unwrapped* (`hasattr(f, "__wrapped__")` → `False`), and does not raise under `warnings.simplefilter("error")`. The current form does raise.
- ✅ **`from strands.vended_tools import shell`** → `ImportError` against published `strands-agents==1.50.2` (`__all__` has `bash`). `file_editor` and `sleep` *do* resolve, so it's specifically `shell` — consistent with your "merged but untagged" note.
- ✅ **Docs claims.** All 5 strandsagents.com links resolve and state their claim; `mcp.slack.com/mcp` returns 401 on a plain GET, which is expected for an OAuth MCP endpoint; the AWS diagram-MCP removal is real (`awslabs/mcp@780e3f1d`, "remove deprecated server (#3096)").
- 🔴 **Not re-run by me:** the full `hatch run prepare`. `tests/test_shell.py::test_command_executor_execute_with_pty` crashes the interpreter in my sandbox on **both** this branch and `main` — pre-existing, matching your note.

</details>

## Questions

❓ **(worth answering before merge)** `calculator`/`cron`/`environment` → "use the `shell` tool vended by strands-agents". The vended shell resolves to `NotASandboxLocalEnvironment` when `Agent(sandbox=...)` is unset, whose own docstring says it runs "directly on the host with no isolation… not a security boundary." So the message trades an AST-allowlisted expression evaluator for model-driven arbitrary host commands by default. You disclose that honestly in both the message and the README — but this repo has spent since ~June 2026 *removing* model-controllable privilege escalation (#525 was on `calculator` itself), and it already ships `code_interpreter` (sandboxed) and `python_repl` (consent-gated), neither deprecated. Is `shell` still the pointer you want, or would `code_interpreter` — or "no replacement needed, sympy is already a dependency" — read better? Your call, not a defect, but the text ships to users the moment this merges.

❓ Since `category=None` (see the `calculator.py` comment) removes the `-W error` breakage with no other observable change, is it worth a small follow-up applying it to #550's `sleep`/`editor`/`shell` too? That would close the gap where #550's body cites "would break consumers running with `-W error`" as its rationale and then ships the decorator that does exactly that.

❓ "This warning becomes an error log in v0.9.0" gives users a log-level change to plan around rather than an action — the SDK's own deprecations name a removal version instead (e.g. "will be removed in Strands SDK 2.0", `strands/tools/loader.py:206`). The real end state is already in your README prose ("this repository will eventually be archived"). Would naming *that* — a Removal/Archive column in place of "Error log" — give users something more concrete?

❓ Given the `strands.vended_tools.shell` import doesn't resolve yet, do you want to ship all 11 now, or split the 3 shell-dependent ones out so the other 8 aren't coupled to a release-sequencing gap that isn't this repo's to close? (`README.md:209`'s copy-pasteable block fails as written today.)

<details>
<summary>Appendix — non-blocking (9 nits + 3 pre-existing)</summary>

**Nits, bundled — none worth its own thread:**
- `getattr(getattr(fn, "_tool_func", None), "__deprecated__", None)` in all 10 marker tests is dead code in both tool shapes — `functools.update_wrapper` (`decorator.py:481`) already copies `__deprecated__` onto the `DecoratedFunctionTool`, and for the legacy shape `fn` *is* the plain function. The left side of the `or` always wins, 10/10.
- `import logging as _logging` / `from strands_tools import X as _mod` inside every new test body is redundant with the file's own top-level import (e.g. `tests/test_batch.py:5`) and diverges from #550's style.
- `diagram.py:24` adds a named `logger` used for exactly one line while 12 pre-existing sites in the same file use the root logger — cosmetic, but it means a future `caplog.at_level(..., logger="strands_tools.diagram")` test only sees the new line.
- `README.md:229` — "it bypassed tracing, metrics, and hooks" is right on tracing and metrics but **wrong on hooks**: `BeforeToolCallEvent`/`AfterToolCallEvent` do still fire under `record_direct_tool_call=False`. The tracing/metrics gap comes from `_ToolCaller.caller` (`strands/tools/_caller.py:118`) calling `ToolExecutor._stream()` and skipping `_stream_with_trace()` — inherent to *any* direct `agent.tool.*` call, not to the flag (confirmed with `record_direct_tool_call=True` too).
- `README.md:190` — the `retrieve` row links `memory/overview/`, but the concrete "a store with only a knowledge base ID is read-only" statement lives on `memory/bedrock-knowledge-base/`, which the sibling `memory` row already links.
- `README.md:231` — "no per-cycle model switching" slightly overclaims: `think.py:347`'s own docstring says "Each cycle uses the same model - mixed model cycles not currently supported," so what's actually lost is cross-*invocation* model choice.
- `think`'s message doesn't mention the loss of tool access inside the reasoning trace (real — `think.py:107-125` passes `filtered_tools` into the nested `Agent`), and "your model provider's reasoning config" only exists for 4 of 15 modules in `strands/models/` (bedrock, anthropic, litellm, sagemaker). An Ollama user gets sent hunting for a knob that isn't there.
- `current_time` → `ContextInjector`: the README covers the timezone loss but not the freshness one — the default `trigger="userTurn"` fires only on a fresh user ask, not on mid-task tool-result turns (`strands/injection/types.py:19-28`), so in a long loop the injected time is stale where a tool call always returned *now*. `trigger="everyTurn"` is the one-word answer. Also the import path is `strands.vended_plugins.context_injector`, and this is the one shell-free message that gives no import at all.
- `diagram` and `rss` are the only two messages with neither an import nor a URL nor an API name — and they're the do-it-yourself cases where a reader needs the most help.
- No push channel for a 14-tool deprecation: `src/strands_tools/__init__.py` is empty, there's no CHANGELOG, and `.github/scripts/group-release-notes.py` has no deprecation bucket — this commit renders under "🚀 Features". A hand-written v0.8.6 note might be worth it.

**Pre-existing, not this PR's problem — flagging, not asking:**
- `tests/test_agent_core_memory.py::test_initialization` asserts `provider.region == os.environ.get("AWS_REGION", "us-west-2")` while `agent_core_memory.py:86,150` hardcodes `DEFAULT_REGION = "us-west-2"` regardless of env — so it fails under any non-default `AWS_REGION`, identically on this branch and on `main`. Confirms your note; happy to file it separately if useful.
- `slack.py:76`'s own usage-example docstring recommends toggling auto-reply "using the environment tool" — this PR deprecates *both* `slack` and `environment`, so that example now points at two deprecated tools. One-line fixup here or a follow-up, your preference.
- `batch`'s `record_direct_tool_call=False` path flattens nested tool results into a dict, which is lossy for non-text content (a nested tool returning an image comes back `str()`-ed). Real and reproduced, but `batch` is being deprecated rather than fixed — noting only because it's in the tool whose README note enumerates what it bypasses.

**Explicitly not findings** (chased and ruled out): tool-spec/schema drift, `agent=` injection breakage, forward-ref/`get_type_hints` resolution through the wrapper (`get_type_hints` walks `__wrapped__`, so PEP-563 annotations still resolve — not a footgun), `warnings` thread-safety under `asyncio.to_thread` (40 concurrent calls across 8 threads, deterministic), `importlib.reload` double-wrapping, and `batch` under default filters. `test_think.py`'s `assert_called_once_with` → `assert_any_call` switch is *correct*, not a weakening — the mocked logger genuinely is called twice now.

</details>

---
🤖 Automated review by `strandly-the-agent` — solid work for a human to approve, not a gate. Findings are evidence-backed and the suggested patches were run; the judgment calls in Questions are yours.

src/strands_tools/batch.py
```diff
@@ -30,12 +30,23 @@
...
+_DEPRECATION_MESSAGE = (
+    "batch is deprecated. This warning becomes an error log in v0.9.0. Concurrent tool execution is now the default "
+    "in the SDK via ConcurrentToolExecutor, so no replacement tool is needed - remove batch and call tools normally. "
+    "See https://strandsagents.com/docs/user-guide/concepts/tools/executors/"
+)
+
+
 TOOL_SPEC = {
     "name": "batch",
     "description": "Invoke multiple other tool calls simultaneously",
```
strandly-the-agent: 🟡 **This commit ships a correction and the claim it corrects, three lines apart.**

The README diff adds "`batch` never actually ran tools in parallel (it looped synchronously)" — which is right, `batch.py:118-126` is a plain `for` loop. But this model-facing description still says "simultaneously", the module docstring (`batch.py:2,4`) still says "in parallel", and `README.md:150` — a row this PR edits — still says "Call multiple other tools in parallel."

The model-facing angle is what makes it worth fixing rather than leaving: a model that reads "simultaneously" and reaches for `batch` *to get concurrency* picks the one tool that can't provide it, when calling the tools normally would have given it real parallelism via `ConcurrentToolExecutor`. Description-only, no behavior change:

```suggestion
    "description": "Invoke multiple other tool calls in sequence (deprecated - does not run them in parallel)",
```

Same edit is worth making at `batch.py:2,4` and `README.md:150`.

src/strands_tools/calculator.py
```diff
@@ -749,7 +756,11 @@ def calculate_series(expr: Any, var: str, point: str, order: int) -> Any:
         raise ValueError(f"Series expansion error: {str(e)}") from e
 
 
+# @deprecated surfaces in IDEs and type checkers; the logger.warning below is what
+# users actually see, since DeprecationWarning raised from inside the SDK's tool
+# invocation path is suppressed by Python's default warning filter.
 @tool
+@deprecated(_DEPRECATION_MESSAGE)
```
strandly-the-agent: 🟡 **`category=None` keeps the static-analysis marker and drops the `-W error` breakage.**

`typing_extensions.deprecated` defaults to `category=DeprecationWarning`, which means it *wraps* the function and raises at call time. Under `python -W error::DeprecationWarning`, `PYTHONWARNINGS=error::DeprecationWarning`, or a downstream project's `filterwarnings = ["error"]`, the SDK's `except Exception` catches that and returns `{"status": "error", ...}` — the tool body never runs. For the 8 `@tool`-shaped tools it's silent (no log line, no traceback); the 3 legacy `TOOL_SPEC` ones at least log an ERROR via `strands/tools/tools.py:281`.

`category=None` is the documented no-runtime-warning form. Verified: `__deprecated__` still set, function returned *unwrapped* (`hasattr(f, "__wrapped__")` → `False`), no raise under `simplefilter("error")`. PEP 702 is explicit that type-checker behavior is unaffected by `category`, so you keep the IDE/mypy strikethrough that's the stated reason for the decorator, and `tool_spec` stays byte-identical.

```suggestion
@deprecated(_DEPRECATION_MESSAGE, category=None)
```

Same one-kwarg change at `batch.py:79`, `cron.py:77`, `current_time.py:24`, `diagram.py:1109`, `environment.py:404`, `memory.py:648`, `retrieve.py:282`, `rss.py:211`, `slack.py:551`, `think.py:196`.

In fairness: the mechanism is inherited from #550, so this isn't something this PR introduces — it widens it from 3 tools to 14 while touching every call site anyway. And the runtime warning wasn't buying much: it's invisible under default filters, and when it does fire it's attributed to `strands/tools/decorator.py:531` or opentelemetry's threading instrumentation rather than the caller's file — the `asyncio.to_thread` hop makes `stacklevel` unfixable from here. The comment above is right that `logger.warning` is the channel that actually reaches people; `category=None` just makes that literally true.

src/strands_tools/slack.py
```diff
@@ -536,7 +544,11 @@ def stop(self):
 socket_handler = SocketModeHandler()
 
 
+# @deprecated surfaces in IDEs and type checkers; the logger.warning below is what
+# users actually see, since DeprecationWarning raised from inside the SDK's tool
+# invocation path is suppressed by Python's default warning filter.
 @tool
+@deprecated(_DEPRECATION_MESSAGE)
```
strandly-the-agent: 🟡 **`slack_send_message` (`slack.py:689`) is registered by this module but gets no deprecation signal.**

`Agent(tools=[slack])` registers both tools. Live, through the public path:

```
agent.tool.slack(action='chat_postMessage', ...)     log_warn=1  DeprecationWarning=1
agent.tool.slack_send_message(channel=.., text=..)   log_warn=0  DeprecationWarning=0
```

It's the only registered tool across all 11 modules that's missed, and the README's Deprecations table names only `slack` — so someone using the simplified wrapper (a reasonable thing to have picked, it's the documented shortcut for the common case) gets no notice at all before v0.9.0 flips to error logs. The same `@deprecated` + first-line `logger.warning` you've applied here should cover it.

README.md
```diff
@@ -117,40 +117,40 @@ Below is a comprehensive table of all available tools, how to use them with an a
...
 | agent_graph | `agent.tool.agent_graph(agents=["agent1", "agent2"], connections=[{"from": "agent1", "to": "agent2"}])` | Create and visualize agent relationship graphs for complex multi-agent systems |
 | graph | `agent.tool.graph(action="create", graph_id="pipeline", topology={"nodes": [...], "edges": [...]})` | Create and manage deterministic DAG-based multi-agent graphs using Strands SDK Graph implementation with per-node model configuration |
-| cron* | `agent.tool.cron(action="schedule", name="task", schedule="0 * * * *", command="backup.sh")` | Schedule and manage recurring tasks with cron job syntax <br> **Does not work on Windows |
-| slack | `agent.tool.slack(action="post_message", channel="general", text="Hello team!")` | Interact with Slack workspace for messaging and monitoring |
+| cron* ⚠️ | `agent.tool.cron(action="schedule", name="task", schedule="0 * * * *", command="backup.sh")` | Schedule and manage recurring tasks with cron job syntax <br> **Does not work on Windows <br> **Deprecated — see [Deprecations](#deprecations)** |
```
strandly-the-agent: 🟡 **This edit breaks the "Deprecated" badge on the `cron*` row specifically.**

The unclosed `**` before "Does not work on Windows" was pre-existing and harmless — it rendered as literal text. Appending `<br> **Deprecated — …**` to that same unclosed run makes markdown bold `**Does not work on Windows <br>**` and leave the Deprecated link *unbolded* with a stray trailing `**`. It's the one row where the badge every other deprecated row displays correctly doesn't render. Two-character fix:

```suggestion
| cron* ⚠️ | `agent.tool.cron(action="schedule", name="task", schedule="0 * * * *", command="backup.sh")` | Schedule and manage recurring tasks with cron job syntax <br> **Does not work on Windows** <br> **Deprecated — see [Deprecations](#deprecations)** |
```

yonib05:
Addressed the review findings in e04ef04 and 698f68b. Two of them were real defects in my own work, and both are now verified by mutation rather than by passing:

**1. The 10 new deprecation tests were tautological.** They emitted the log record themselves and then asserted on `caplog`, so they only proved the module constant was a string. Confirmed by deleting `logger.warning` from all 11 modules — every test stayed green.

They now invoke the tool. `TypeError` is re-raised rather than swallowed, because a signature mismatch means the tool was never entered and the warning would be missing for the wrong reason — which is exactly how the `think` test was passing an incomplete argument list. Other exceptions stay swallowed, since several tools legitimately fail without credentials or optional deps and the warning fires before that work.

Re-ran the mutation: **removing `logger.warning` from the 11 modules now fails 11 tests.** Also added the `calculator` pair, which was missing entirely.

**2. The security assertion was weakened, and the slack was larger than it looked.** `assert_not_called()` had become `call_count - calls_before <= 1` to tolerate stdlib `logging` calling `os.getpid()` while formatting a record. But that budget is exactly what a minimal escape needs, and for the `builtins.open` payload logging contributes *zero*, so the tolerance was pure slack. `calls_before` was also always 0, making the subtraction dead code.

Fixed by switching the payloads to `os.getppid`, which `logging` never touches — verified — so `assert_not_called()` is back with no tolerance. Confirmed strict by injecting a single `getppid()` call into the tool body: the test now fails, where the old bound would have passed it.

**Also fixed:**
- `slack_send_message` was the only registered tool across the 11 modules getting no deprecation signal at all.
- Dropped the `_tool_func` fallback in the marker lookup — dead for all 11 tools, and it would have kept the test green if the SDK ever stopped forwarding `__deprecated__`.
- The `cron` README row had a pre-existing unterminated `**`; appending the notice made it render with literal asterisks.
- `batch`'s own description claimed "simultaneously" / "in parallel" while this PR's README note says it never was. The body is a plain synchronous `for` loop, so I corrected the description rather than the note.

`hatch run prepare`: **1330 passed, 1 failed** — `test_agent_core_memory.py::test_initialization`, which asserts a hardcoded region against local `AWS_REGION` and fails identically on `main`.

**One item I have not changed, because it needs a maintainer call:** `@deprecated(...)` without `category=None` means consumers running `-W error` get a raised `DeprecationWarning` instead of a result. Passing `category=None` keeps the IDE/type-checker marker while silencing the runtime raise — verified. But that decision also applies to the three tools already merged in #550, so it seems better made once, deliberately, than quietly here. Happy to apply it to both if you'd like.

tests/test_calculator.py
yonib05: Fixed — payloads switched to os.getppid (logging never calls it), so assert_not_called() is back with no tolerance. Verified strict by injecting a getppid() call into the tool body: the test now fails.

tests/test_batch.py
yonib05: Fixed — the tests now invoke the tool. Verified by mutation: deleting logger.warning from all 11 modules now fails 11 tests.

src/strands_tools/batch.py
yonib05: Fixed — corrected the docstring, TOOL_SPEC description, and README row to say the invocations run in sequence.

src/strands_tools/slack.py
yonib05: Fixed — slack_send_message now carries the same decorator and log line.

README.md
yonib05: Fixed — closed the pre-existing unterminated ** in the cron row.

src/strands_tools/calculator.py
yonib05: Verified the mechanism — `category=None` does keep `__deprecated__` while dropping the `-W error` raise. Not applying it here because the same decision applies to the three tools already merged in #550, so it should be made once for both rather than diverging. Flagged for a maintainer call.

yonib05:
@strandly-the-agent all six findings are addressed and the threads are resolved — please re-review.

The two substantive ones are verified by mutation rather than by passing: deleting `logger.warning` from all 11 modules now fails 11 tests (previously all green), and injecting a single `getppid()` call into the calculator body now fails the security test (previously passed under the `<= 1` bound).

One left open deliberately: `category=None`. Same decision applies to the three tools merged in #550, so it seems better made once for both than diverging here.

README.md
yonib05: Fixed — reworded; the section now leads with why the capability moved rather than announcing removal.

README.md
yonib05: Fixed — reframed around what to do instead, with the trade-off stated plainly rather than as a warning.

README.md
yonib05: Fixed — dropped the editorialising about what AWS concluded.

README.md
yonib05: Fixed — now links https://docs.slack.dev/ai/mcp-server/ (docs rather than the bare endpoint), and the table row too.

README.md
yonib05: Fixed — dropped the AWS naming; the cron row now says "a managed scheduler" and the diagram note no longer editorialises.

README.md
yonib05: Fixed — rewrote the whole section. The preamble now explains that the SDK had no built-in way to do these things when the package started and much of it is now native or vendor-maintained, so the reason is stated up front. Per-tool notes are reframed around what to do instead, with trade-offs stated plainly rather than as "you also lose X", and each native-capability entry links its docs page.

yonib05:
@strandly-the-agent all 12 threads are addressed and resolved — ready for re-review.

The README messaging is rewritten per the tone feedback: it now explains *why* each capability moved rather than announcing removal, links the docs page for every native-capability entry, and drops the AWS editorialising. All 7 links verified 200.

README.md
yonib05: Over-corrected on this one — I read it as "no AWS references" and genericised the cron row to "a managed scheduler", which lost a useful pointer. Restored Amazon EventBridge Scheduler; what's removed is the sentence narrating what AWS decided about their own diagram MCP server, which is what didn't belong.

yonib05:
@strandly-the-agent can you review this PR?

tests/test_think.py
```diff
@@ -642,3 +645,34 @@ def test_think_tool_recursion_prevention_multiple_cycles():
...
+
+def test_think_logs_deprecation_warning(caplog):
+    """Invoking the tool logs a deprecation warning naming the migration path."""
+    import logging as _logging
+
+    from strands_tools import think as _mod
+
+    with caplog.at_level(_logging.WARNING, logger="strands_tools.think"):
+        try:
+            _mod.think(thought="t", cycle_count=1, system_prompt="s")
```
strandly-the-agent: 🔴 **This now performs a live Bedrock model invocation on every test run.**

Making the tests actually invoke the tool was the right fix, but the invocations are unmocked, so four of the eleven do real I/O. Socket-level spy over the 11 log tests:

```
test_think_logs_deprecation_warning     -> 169.254.169.254:80 (IMDS) + 54.212.242.79:443
test_retrieve_logs_deprecation_warning  -> 169.254.169.254:80 (IMDS)
test_cron_logs_deprecation_warning      -> subprocess: crontab -l   (reads the dev's real crontab)
test_memory_logs_deprecation_warning    -> AWS endpoint once STRANDS_KNOWLEDGE_BASE_ID is set
```

Calling it directly shows how far it gets:

```
Error in think tool: An error occurred (AccessDeniedException) when calling the ConverseStream
operation: User: arn:aws:sts::...:assumed-role/... is not authorized to perform:
bedrock:InvokeModelWithResponseStream on resource: ...claude-sonnet-4-6
```

That's an **IAM denial, not a code guard** — on a contributor's laptop with `~/.aws/credentials`, or CI with a Bedrock-capable role, it succeeds and bills a real model call. Same spy against `fab13ea`: zero I/O for all of these, so it's new in this round. And `except Exception: pass` means nobody ever notices.

The file already has the pattern — `test_think_tool_direct` mocks the nested Agent:

```suggestion
        try:
            with patch("strands_tools.think.Agent") as mock_agent_class:
                mock_agent_class.return_value.return_value = AgentResult(
                    message={"content": [{"text": "mock"}]},
                    stop_reason="end_turn",
                    metrics=None,
                    state=MagicMock(),
                )
                _mod.think(thought="t", cycle_count=1, system_prompt="s", agent=None)
```

The other three want the fixtures already sitting in their own files: `mock_subprocess` (`test_cron.py:20`), `mock_memory_service_client` (`test_memory/test_memory.py:14-27`), `mock_boto3_client` (`test_retrieve.py`, already used by `test_retrieve_tool_direct`). A unit test whose entire subject is the function's first line shouldn't need the network to run.

tests/test_cron.py
```diff
@@ -429,3 +429,34 @@ def test_consent_granted_allows_write(mock_subprocess, monkeypatch):
...
+    from strands_tools import cron as _mod
+
+    with caplog.at_level(_logging.WARNING, logger="strands_tools.cron"):
+        try:
+            _mod.cron(action="list")
+        except TypeError:
+            # A signature mismatch means the tool was never entered, so the
+            # warning would be missing for the wrong reason - fail loudly.
+            raise
+        except Exception:
```
strandly-the-agent: 🟡 **This swallow is dead code for all 11 tools — and where it does fire, it destroys the cause.**

I checked what each of the 11 invocations actually does on failure: **none of them raises.** All 11 tools already wrap their bodies in broad handlers and return a `status: "error"` dict instead. So the stated rationale ("tools legitimately fail without credentials or optional deps") doesn't hold for any of the eleven — `cron(action="list")` returns `{'status': 'error', ... "No such file or directory: 'crontab'"}` rather than raising.

Where it *does* fire is the case you deferred. Under `-W error::DeprecationWarning`, `warnings.warn` raises inside `@deprecated` before the body is entered, and this `except Exception: pass` eats it:

```
$ pytest <the 11 log tests> -W error::DeprecationWarning -q
E  AssertionError: assert 'DEPRECATION WARNING' in ''      x10
FAILED tests/test_calculator.py::... - DeprecationWarning: calculator is deprecated...  <- real cause
FAILED tests/test_cron.py::...       - AssertionError                                   <- cause destroyed
```

Your new `calculator` test is the only one without the try/except, and it's the only one that reports the truth. So the swallow is actively hiding the `category=None` issue from whoever eventually investigates it.

Dropping both handlers leaves the tests passing and makes any future breakage report itself. If you'd rather keep a guard, narrowing it to the specific exception a tool can actually raise beats a bare `Exception`.

tests/test_cron.py
```diff
@@ -429,3 +429,34 @@ def test_consent_granted_allows_write(mock_subprocess, monkeypatch):
...
+            # The tool may still fail without credentials or optional deps; the
+            # deprecation warning is logged before any of that work happens.
```
strandly-the-agent: 🟡 **The tool-specific migration anchor was dropped from 10 of the 11 log tests.**

This test used to assert `"strands.vended_tools import shell" in caplog.text`; now it checks only `"DEPRECATION WARNING"` and `"becomes an error log in v0.9.0"` — both generic across all 14 deprecated tools. So it no longer proves *which* message was logged: swap `cron`'s `logger.warning` to emit `batch`'s constant and this stays green.

It's also inconsistent within the same commit — your new `calculator` log test kept its anchor (`assert "strands.vended_tools import shell" in caplog.text`), and that's the right shape. I restored the assertion to all 10 and they pass (`11 passed`), so nothing was in the way of keeping them:

```suggestion
    assert "DEPRECATION WARNING" in caplog.text
    assert "becomes an error log in v0.9.0" in caplog.text
    assert "strands.vended_tools import shell" in caplog.text
```

The marker test still anchors the constant tool-specifically, so this isn't a hole — just a cheaper, narrower log test than the one you wrote for `calculator`.

src/strands_tools/slack.py
```diff
@@ -673,6 +687,7 @@ def slack(action: str, parameters: Dict[str, Any] = None, agent=None) -> str:
 
 
 @tool
+@deprecated(_DEPRECATION_MESSAGE)
```
strandly-the-agent: 🟡 **`slack_send_message`'s deprecation is the only one with no test.**

The fix itself is right. But it's the sole new source code in this round and the only one of the **12** deprecated callables with nothing guarding it — note your mutation claim covers "11 modules", and `slack.py` has two decorated callables.

Two mutations, each on a clean copy:

```
$ # delete slack_send_message's logger.warning (keep slack()'s)
$ pytest tests/test_slack/ -q   ->  32 passed, 3 skipped
$ # also remove its @deprecated decorator
$ pytest tests/test_slack/ -q   ->  32 passed, 3 skipped
```

Identical to baseline both times — so if either regresses, nothing catches it. `tests/test_slack/test_slack.py` already has three tests calling `slack_send_message(channel="test_channel", text="test message")` under a mocked client (lines 151-201), so the deprecation pair is a copy of the `slack` pair with that call substituted.

README.md
```diff
@@ -183,6 +188,17 @@ exists as well.
...
+| `think` | native extended thinking via model reasoning config ([docs](https://strandsagents.com/docs/user-guide/concepts/model-providers/amazon-bedrock/)) | v0.8.6 | v0.9.0 |
+| `current_time` | `ContextInjector` ([docs](https://strandsagents.com/docs/user-guide/concepts/plugins/context-injector/)) | v0.8.6 | v0.9.0 |
+| `memory` | `MemoryManager` + `BedrockKnowledgeBaseStore` ([docs](https://strandsagents.com/docs/user-guide/concepts/memory/bedrock-knowledge-base/)) | v0.8.6 | v0.9.0 |
+| `retrieve` | `MemoryManager` + `BedrockKnowledgeBaseStore(writable=False)` ([docs](https://strandsagents.com/docs/user-guide/concepts/memory/overview/)) | v0.8.6 | v0.9.0 |
+| `calculator` | `from strands.vended_tools import shell` (run `python3 -c` with sympy) | v0.8.6 | v0.9.0 |
+| `cron` | `from strands.vended_tools import shell` (manage `crontab`), or Amazon EventBridge Scheduler | v0.8.6 | v0.9.0 |
+| `environment` | `from strands.vended_tools import shell` (inspect only — see notes) | v0.8.6 | v0.9.0 |
+| `slack` | [official Slack MCP server](https://docs.slack.dev/ai/mcp-server/); `slack_bolt` for Socket Mode | v0.8.6 | v0.9.0 |
```
strandly-the-agent, line 204: 🟡 **This rewrite moved the Slack URL, but the runtime message still points at the old one.**

This row and the note at `README.md:269` now say `docs.slack.dev/ai/mcp-server/` — a better pointer, and it returns 200. But `src/strands_tools/slack.py:142`'s `_DEPRECATION_MESSAGE` still reads:

> `use the official Slack MCP server at https://mcp.slack.com/mcp`

That's the string a user actually *sees*, and `mcp.slack.com/mcp` returns 401 to a browser (correct for an OAuth MCP endpoint, but not a page anyone can read for setup instructions). Round 1 found every row's note agreeing with its in-code message; this is the one place the rewrite broke that.

Worth updating `slack.py:142` to the docs URL — and note `tests/test_slack/test_slack.py:406` asserts `"mcp.slack.com" in marker`, so that anchor needs to move with it.

src/strands_tools/batch.py
```diff
@@ -1,7 +1,8 @@
 """
 Batch Tool for Parallel Tool Invocation
```
strandly-the-agent: 🟡 **"Parallel" survives in the two most-read spots.**

Lines 4-5 just below correctly say the invocations run sequentially, so the module header now contradicts its own second sentence. Same in the feature list at `README.md:60`: *"🔄 **Multiple tools in Parallel** - Call multiple other tools at the same time in parallel with Batch Tool"*, which contradicts the migration note at `README.md:229`.

`TOOL_SPEC["description"]` and the tools-table row are both correctly fixed — these two are just leftovers from the same sweep.

```suggestion
Batch Tool for Multi-Tool Invocation
```

strandly-the-agent (commented):
Re-reviewed at `25d3b07`. **Both of your mutation claims hold — I reproduced them independently.** Round-1's six findings are genuinely fixed; the round-2 diff introduced one new 🔴 and left a few loose ends. Not re-raising anything from the 12 resolved threads.

| Your claim | My result |
|---|---|
| deleting `logger.warning` from all 11 modules now fails 11 tests | ✅ **11 failed** (was 11 green) — zero surviving mutants |
| injecting one `getppid()` into the calculator body fails the security test | ✅ **fails** — and injecting one `open('/dev/null')` fails the `builtins.open` param too, which the old `<= 1` bound let through |

The `os.getppid` switch is better than the fix I suggested: it removes the interference instead of masking it. Confirmed `LogRecord.__init__` references `getpid` and never `getppid` (0 `getppid` calls after a successful `calculator()` that logs a record, vs 1 `getpid`).

**The one new problem:** making the tests invoke the tools left them unmocked, so four of them now do real I/O — `tests/test_think.py:658` reaches Bedrock `ConverseStream`. Details inline.

## Your `category=None` question — answered

Your reasoning is right and I'd defer it too: applying it here and not to `sleep`/`editor`/`shell` would leave 14 tools with two different runtime contracts, which is worse than either choice made uniformly. So — **don't apply it in this PR.** My concrete suggestion: land this, then one follow-up that sets `category=None` on all 14 call sites at once. Worth doing rather than dropping, because it's the difference between a consumer with `filterwarnings = ["error"]` getting a result and getting a silent `status: "error"`.

One wrinkle that makes it more urgent than it looks: the new `except Exception: pass` in the log tests **swallows exactly this failure mode** — under `-W error::DeprecationWarning` 10 of the 11 tests fail with `assert 'DEPRECATION WARNING' in ''` and the real `DeprecationWarning` cause is destroyed. Your new `calculator` test, the only one without the try/except, is the only one that reports the truth. See the `test_cron.py` comment.

<details>
<summary>✅ Verification ledger — what I ran at 25d3b07</summary>

- ✅ `git diff --stat 698f68b 25d3b07` → **README.md only** (58+/39−), so every code/test finding I verified at `698f68b` still stands verbatim; I re-anchored all line numbers against head.
- ✅ **Mutation A:** deleted `logger.warning` from all 11 modules → `11 failed`. Your claim, reproduced.
- ✅ **Mutation B:** injected one `os.getppid()` after the deprecation log in `calculator` → `2 failed, 1 passed` (both `getppid` params). Separately injected one `open('/dev/null')` → the `builtins.open` param fails. Strictly stronger than round 1.
- ✅ **Mutation C:** deleted `slack_send_message`'s `logger.warning`, then its `@deprecated` → `tests/test_slack/` stayed at **32 passed, 3 skipped** both times. Untested.
- ✅ **Socket-level I/O spy** over the 11 log tests: `test_think` → `169.254.169.254:80` (IMDS) + `54.212.242.79:443`; `test_retrieve` → IMDS; `test_cron` → `crontab -l` subprocess; `test_memory` → no network here only because no KB id is set. Direct call confirms `think` gets an `AccessDeniedException` from `bedrock:InvokeModelWithResponseStream` — an IAM denial, not a code guard.
- ✅ **All 7 README links return 200**, including the two new ones and the changed Slack one: `tools/executors/`, `model-providers/amazon-bedrock/`, `plugins/context-injector/`, `memory/bedrock-knowledge-base/`, `memory/overview/`, `memory/overview/#custom-stores`, `github.com/modelcontextprotocol/servers/tree/main/src/time`, `docs.slack.dev/ai/mcp-server/`. Your claim checks out.
- ✅ The rewritten `batch` note now says the SDK executor *keeps* tracing/metrics/hooks intact — that's accurate, and it resolves the incorrect "batch bypassed … hooks" claim from round 1. The AWS editorialising is gone as asked.
- ✅ `slack_send_message`'s new decorator did **not** change its generated `tool_spec`; dropping the `_tool_func` fallback still resolves `__deprecated__` directly on all 12 callables, both tool shapes.
- ✅ The `cron*` README row now renders two `<strong>` spans with no stray `**` (python-markdown + tables).
- 🔴 Not re-run: full `hatch run prepare`. Same pre-existing `test_agent_core_memory.py::test_initialization` region failure on both head and `main`, matching your note.

</details>

<details>
<summary>Appendix — non-blocking (3)</summary>

- **`TypeError` is inert for the 3 legacy `TOOL_SPEC` tools.** `batch`/`environment`/`retrieve` take `(tool: ToolUse, **kwargs)`, so a signature mismatch can't happen — the guard only does work for the 8 `@tool` tools. Renaming a key inside `environment.py` makes it raise `KeyError`, and the log test still passes (its 30 sibling tests catch it, which is why this is a nit).
- **`diagram(diagram_type="graph", nodes=[], edges=[])` is rejected at validation** (`"Error: 'nodes' parameter is required for basic diagrams"`), so that invocation exercises almost none of the tool. It does still prove the warning fires, which is the point — just noting it's the thinnest of the 11.
- **The security test has no positive control:** nothing asserts `getppid` is reachable when the guard is bypassed, so a typo'd payload string would pass silently. A companion assertion that the sentinel *does* fire when the AST guard is deliberately bypassed would close that.

Also checked and explicitly **not** findings: the `getppid` sentinel is airtight under `--cov`, xdist, `--log-file`, and forced `logMultiprocessing`/`logProcesses` (zero `getppid` calls suite-wide); asserts sitting inside `mock.patch("builtins.open")` are safe because `__exit__` restores `open` before pytest renders a traceback; no parametrize coverage regressed; `/tmp/strands_rss_feeds` and `./slack_events/` are created at module import and pre-date this PR.

</details>

---
🤖 Round-2 review by `strandly-the-agent`. The two substantive round-1 fixes are verified by mutation, not by assertion. The 🔴 is a hermeticity regression with a small fix, not a design problem.

strandly-the-agent:
Already done — the re-review is [here](https://github.com/strands-agents/tools/pull/566#pullrequestreview-4860571796), posted 02:51:42Z against `25d3b07`, which is still the current head (last push 02:34:58Z). Its six inline threads are open and unanswered, so nothing is waiting on me.

Flagging a loop: this is the sixth `@strandly-the-agent` mention here since 02:19Z with no push in between, and each one kicks off a full review run. If you pushed something you expected me to pick up, it didn't land — `git log` still ends at `25d3b07`.

One thing that *is* new since I reviewed: **CI is green at `25d3b07`** (19/19, unit tests on 3.10–3.13 across linux/windows/macOS). Worth saying explicitly that this doesn't contradict the 🔴 — the hermeticity finding passes CI *by construction*, because `except Exception: pass` swallows whatever the unmocked call does. On a runner without AWS credentials `test_think` fails fast and is swallowed; on a machine that has them it's a real `ConverseStream` call and is also swallowed. Green CI is exactly what that finding predicts.

<details>
<summary>TL;DR of the review, if it's easier than opening it</summary>

Both your mutation claims verified true. Your `os.getppid` switch is better than the fix I'd suggested. All 7 README links 200, and the reworded `batch` note fixes the wrong "bypassed hooks" claim from round 1.

- 🔴 `tests/test_think.py:658` — the unmocked invocations do real I/O; `think` reaches Bedrock `ConverseStream` (`AccessDeniedException` is an IAM denial, not a code guard). `test_cron` runs `crontab -l`, `test_retrieve` hits IMDS. Fix: the mocks already in each file.
- 🟡 `tests/test_cron.py:447` — `except Exception: pass` is dead code (no tool raises) and destroys the real cause under `-W error`.
- 🟡 `tests/test_cron.py:449` — 10 of 11 log tests lost their tool-specific migration anchor; your new `calculator` test kept its.
- 🟡 `src/strands_tools/slack.py:690` — `slack_send_message`'s deprecation is untested; both mutations leave `tests/test_slack/` at 32 passed.
- 🟡 `README.md:199` — the rewrite moved the Slack URL to `docs.slack.dev`, but `slack.py:142` still says `mcp.slack.com/mcp`.
- 🟡 `src/strands_tools/batch.py:2` — "Parallel" survives in the docstring title and `README.md:60`.

And on `category=None`: agreed, don't do it here — one follow-up across all 14 sites including #550's three.

</details>

tests/test_think.py
yonib05: Fixed — think and cron are now mocked at the I/O boundary. Verified zero sockets and zero subprocess calls afterwards; the deprecation suite drops 47s to 13s.

tests/test_cron.py
yonib05: Fixed — the try/except is gone. You were right that none of the eleven raises, so it was dead code that could only hide a failure.

tests/test_cron.py
yonib05: Fixed — per-tool anchor restored to all 11. Verified: swapping cron's message for batch's now fails.

src/strands_tools/slack.py
yonib05: Fixed — added both tests for slack_send_message. Mutation now fails 12, up from 11.

README.md
yonib05: Fixed — slack.py's _DEPRECATION_MESSAGE now points at docs.slack.dev/ai/mcp-server/, matching the README.

src/strands_tools/batch.py
yonib05: Fixed — module header and README feature list no longer say parallel.

yonib05:
@strandly-the-agent all 6 addressed and resolved — please re-review. The I/O finding was the important one: verified zero sockets and zero subprocess calls after mocking, and the suite went 47s to 13s. Mutation now fails 12 tests, up from 11.

tests/test_retrieve.py
```diff
@@ -746,3 +746,26 @@ def test_retrieve_via_agent_with_enable_metadata(agent, mock_boto3_client):
...
+
+
+def test_retrieve_logs_deprecation_warning(caplog):
+    """Invoking the tool logs a deprecation warning naming its migration path."""
+    import logging as _logging
+
+    from strands_tools import retrieve as _mod
+
+    with caplog.at_level(_logging.WARNING, logger="strands_tools.retrieve"):
+        _mod.retrieve({"toolUseId": "t", "input": {"text": "q"}})
```
strandly-the-agent, line 763: 🟡 **Non-blocking — two of the four still do I/O; `think` and `cron` are genuinely fixed.**

The mocks you added work: `think` and `cron` now show zero sockets and zero subprocess calls, confirmed both in the full selection and in isolation. But `retrieve` and `memory` only had the `try/except` swapped for the anchor, and they're the two backed by boto3:

```
$ pytest tests/test_retrieve.py::test_retrieve_logs_deprecation_warning -q
1 passed
OUTBOUND: ['169.254.169.254:80']            # IMDS, unconditional

$ STRANDS_KNOWLEDGE_BASE_ID=ABCDEF1234 AWS_REGION=us-west-2 \
    pytest tests/test_memory/test_memory.py::test_memory_logs_deprecation_warning -q
1 passed
OUTBOUND: ['169.254.169.254:80', 'bedrock-agent.us-west-2.amazonaws.com:443']
```

`retrieve.py:363` constructs `boto3.client("bedrock-agent-runtime", ...)`, which kicks off credential resolution before botocore's own parameter validation rejects the call — so the IMDS probe happens on every run. `memory` looked clean in your check only because `STRANDS_KNOWLEDGE_BASE_ID` was unset; with it set it goes all the way to the service. That's the env var anyone working on the memory tool would have exported, which is why I'd rather it not depend on that.

Low severity — an IMDS probe costs a timeout, not money or data, and this is where the residual latency in your 47s baseline lives. But both files already have the fixture:

```suggestion
    with mock.patch("strands_tools.retrieve.boto3.client"), caplog.at_level(_logging.WARNING, logger="strands_tools.retrieve"):
        _mod.retrieve({"toolUseId": "t", "input": {"text": "q"}})
```

(`mock_boto3_client` at `tests/test_retrieve.py:23` is the same patch, already used by `test_retrieve_tool_direct`; `mock_memory_service_client` at `tests/test_memory/test_memory.py:22` is the equivalent for `memory`.) Worth folding in whenever you next touch this — not worth a round trip on its own.

strandly-the-agent (approved):
Re-reviewed at `d1c127f`. **All six round-2 findings are addressed, and both of your new claims verified true.** One small leftover on the I/O finding — non-blocking, one line each. Approving.

| Your claim | My result |
|---|---|
| mutation now fails 12 tests, up from 11 | ✅ **12 failed, 12 passed** — removed all 12 `logger.warning` call sites; the new `slack_send_message` pair is what closes the 11→12 gap |
| per-tool anchor restored; swapping cron's message for batch's now fails | ✅ **2 failed** — swapped `cron`'s constant for batch-style text and both its tests caught it |
| zero sockets / zero subprocess after mocking | ⚠️ **true for `think` and `cron`** (both now zero) — but `test_retrieve` still reaches IMDS, and `test_memory` reaches `bedrock-agent.us-west-2.amazonaws.com:443` once `STRANDS_KNOWLEDGE_BASE_ID` is set. Details inline. |
| suite 47s → 13s | Can't reproduce, and that's expected — see below |

On the timing: I measured **2.12s → 2.01s** for the same test selection across the two heads. That's not a contradiction of your numbers, it's a different environment — my sandbox has a fast-responding IMDS, so the credential probe and the Bedrock call that dominate your 47s cost me almost nothing here. The mechanism you removed is real and I confirmed it's gone for `think`/`cron`; I just can't put your figure on it.

The `try/except` removal is right — you were correct to drop it rather than narrow it. And thanks for taking the `-W error` reasoning on board even though `category=None` itself stays deferred; that's still the one open item, by agreement, for a follow-up across all 14 sites including #550's three.

<details>
<summary>✅ Verification ledger — what I ran at d1c127f</summary>

- ✅ `git diff --name-only 25d3b07 d1c127f` → README.md, `batch.py`, `slack.py` + 11 test files. Reviewed the incremental diff only; the 8 resolved threads are closed and I'm not re-raising them.
- ✅ **Mutation (12 sites):** stripped every `logger.warning("DEPRECATION WARNING: %s", ...)` from the 11 modules → `12 failed, 12 passed`. Your 11→12 claim, reproduced.
- ✅ **Anchor swap:** replaced `cron`'s `_DEPRECATION_MESSAGE` with batch-style text → `test_cron_logs_deprecation_warning` **and** the marker test both fail. The per-tool anchor is genuinely load-bearing again.
- ✅ **Per-test I/O attribution** (socket `connect`/`create_connection`/`getaddrinfo` + `subprocess.run`/`Popen`/`check_output`/`call` spies, attributed by pytest nodeid) across all 24 deprecation tests: only `test_retrieve_logs_deprecation_warning` shows a hit. `think` and `cron` are clean — confirmed in isolation too.
- ✅ `test_slack_send_message_logs_deprecation_warning` / `..._is_marked_deprecated_for_static_analysis` both present and both catch their mutation; the marker anchor moved to `docs.slack.dev` in step with `slack.py:141`.
- ✅ `slack.py`'s `_DEPRECATION_MESSAGE` now names `https://docs.slack.dev/ai/mcp-server/`, matching the README row and note — the round-2 divergence is closed.
- ✅ `batch.py:2` is now "Batch Tool for Multi-Tool Invocation" and `README.md:60` reads "Multiple Tools per Turn"; no "parallel"/"simultaneous" claim survives anywhere (`grep -i` clean).
- ✅ All 11 touched test files, full run: **398 passed, 3 skipped in 9.46s**. `ruff check` clean; `ruff format --check` clean on all 13 changed Python files (an earlier "1 file would be reformatted" was my own error — I'd passed README.md to the formatter).
- 🔴 Not re-run: full `hatch run prepare`. Note `test-format` isn't in the `prepare` chain (`format`, `lint`, `test-lint`, `test`), so formatting drift wouldn't surface there — not an issue on this diff, just worth knowing.

</details>

---
🤖 Round-3 review by `strandly-the-agent`. Approving as a signal that I found no blockers, not as a merge gate — a maintainer still owns that call, and `category=None` is theirs to schedule.

src/strands_tools/batch.py
```diff
@@ -30,15 +31,26 @@
...
+    "batch is deprecated. This warning becomes an error log in v0.9.0. Concurrent tool execution is now the default "
+    "in the SDK via ConcurrentToolExecutor, so no replacement tool is needed - remove batch and call tools normally. "
+    "See https://strandsagents.com/docs/user-guide/concepts/tools/executors/"
+)
+
+
 TOOL_SPEC = {
     "name": "batch",
-    "description": "Invoke multiple other tool calls simultaneously",
+    "description": "Invoke multiple other tool calls in sequence from one request",
```
mkmeral, line 53: nit: we should not change descriptions or other stuff

src/strands_tools/calculator.py
```diff
@@ -94,11 +94,18 @@
...
     standard_transformations,
 )
+from typing_extensions import deprecated
 
 from strands_tools.utils import console_util
 
 logger = logging.getLogger(__name__)
 
+_DEPRECATION_MESSAGE = (
+    "calculator is deprecated. This warning becomes an error log in v0.9.0. Migration path: use the shell tool vended "
```
mkmeral: I don't want to call this a "migration path:". Just say, you can look into using shell or something. Calling it as migration path makes me think I can safely move there

mkmeral: same with cron and other shell recommended tools. the whole security posture changes so it's not a direct migration

src/strands_tools/cron.py
```diff
@@ -11,12 +11,19 @@
...
 from strands import tool
+from typing_extensions import deprecated
 
 from strands_tools.utils import console_util
 from strands_tools.utils.user_input import get_user_input
 
 logger = logging.getLogger(__name__)
 
+_DEPRECATION_MESSAGE = (
+    "cron is deprecated. This warning becomes an error log in v0.9.0. Migration path: use the shell tool vended by "
```
mkmeral: nit: fwiw i want cron in sdk at some point for ambient agents and looping. but not a prio rn

mkmeral (commented):
the main issue i have is shell being presented as migration path directly. i'd want the message to be a bit more nuanced

src/strands_tools/calculator.py
yonib05: Valid point suggestions? I will think of some now

I agree about cron, but lets take it one step at a time and re-release some high quality versions optimized for the sdk or even introduce it as a native functionality seems reasonable.

src/strands_tools/batch.py
yonib05: Ack, it's a fix to the actual implementation, and the migration to the native behavior in sdk actually makes this better, but I hear you. Rather keep it since, IMO, it's a bug not doing what it says it is supposed to.

mkmeral (dismissed):

mkmeral (approved):

PR #585, opened 2026-08-25
fix(calculator): reject cls= reroute in symbols()

poshinchen:
## Summary

The calculator tool validates every expression against an AST allowlist before evaluation. The validator trusted a string literal whenever it appeared as a positional argument to a safe string constructor (`Symbol`, `symbols`, `Rational`, `Integer`, `Float`), but it never inspected the call's keyword arguments.

`symbols()` accepts a keyword-only `cls=` argument that changes which constructor it applies to the parsed name. Passing `cls=N` turns `symbols("<expr>", cls=N)` into `N("<expr>")`, which re-parses the string through SymPy's `sympify`. `sympify` evaluates with full builtins, so a crafted string escapes the restricted namespace and runs arbitrary code on the host.

This expression passed validation and executed before the fix:

```
symbols("<payload>", cls=N)
```

## Fix

A string positional argument is now trusted only when every keyword on the call is a boolean assumption flag (for example `positive=True`, `real=True`) and there is no `**kwargs` unpacking. Any other keyword value, including `cls=<constructor>`, drops the call out of the trusted set, so its string literals are rejected with the existing validation error.

This closes the `cls=` route and any other keyword that could reroute string parsing into `sympify`, while keeping the documented usage working: `Symbol('x')`, `symbols('x y')`, `Rational('1/3')`, and assumption keywords like `Symbol('x', positive=True)`.

## Tests

- Reject the `cls=` reroute, including a mock that confirms no code executes.
- Confirm assumption keywords still pass validation.
- All existing calculator tests remain green.

## Note

The calculator tool is deprecated, but it is still shipped and executable in current releases, so the vulnerable code path remains reachable until users migrate off it. This patch protects those users; it is independent of the deprecation.

### DIFF
--- a/src/strands_tools/calculator.py
+++ b/src/strands_tools/calculator.py
@@ -20,6 +20,11 @@ Security:
     name or numeric literal rather than through sympify (Symbol, symbols, Rational,
     Integer, Float); for example Symbol('x') and Rational('1/3') remain supported.
 
+    A string positional argument to those constructors is trusted only when the call
+    carries nothing but boolean assumption keywords (for example Symbol('x', positive=True)).
+    A keyword that reroutes parsing, most importantly symbols('...', cls=N), is rejected,
+    because cls=N makes symbols apply N (and therefore sympify) to the string.
+
 Key Features:
 1. Expression Evaluation:
    • Basic arithmetic operations (addition, multiplication, etc.)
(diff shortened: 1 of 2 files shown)


liramon2:
@strandly-the-agent Is this safe to merge?

liramon2 (approved):

poshinchen:
@strandly-the-agent review the PR

